PR #1, opened 2015-05-04
Adds support for libvirt rbd volume backend #3285

clsacramento:
Uses a file "ceph.conf" to get CEPH cluster info and decide if it is RBD disk instead of a file one.

Still need to figure out how to pass this info instead of reading this file...

Help please!?

### DIFF
--- a/lib/fog/libvirt/models/compute/templates/ceph.conf
+++ b/lib/fog/libvirt/models/compute/templates/ceph.conf
@@ -1,0 +1,5 @@
+monitor=192.168.100.50,192.168.100.51,192.168.100.52
+port=6789
+libvirt_ceph_pool=foreman-images
+auth_username=foreman
+auth_uuid=8e1050ae-a53c-4280-bc48-8abe0eebafb1
--- a/lib/fog/libvirt/models/compute/templates/server.xml.erb
+++ b/lib/fog/libvirt/models/compute/templates/server.xml.erb
@@ -15,13 +15,35 @@
   </features>
   <clock offset='utc'/>
   <devices>
-<% volumes.each do |vol| -%>
-    <disk type='file' device='disk'>
-      <driver name='qemu' type='<%= vol.format_type %>'/>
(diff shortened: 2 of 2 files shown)


strzibny:
Hi, I am not that familiar with Ceph, can you elaborate more on what you are trying to achieve and how to properly test this behaviour?

clsacramento:
Hi,

I am trying to deploy a VM wich its disk is not a file on the hypervisor
but is on Ceph instead (the block device is directly accessed on the VM, it
is not mounted on the host). This means the volume type in Libvirt XML must
be "rbd", not "file" as in the template, and there are also few
informations about hte Ceph cluster to fill the XML with.

The best way to test would be check that you can still deploy VMs on mode
file, and test that it is now possible to deploy on Ceph. Unfortunately,
you will need to have access to a Ceph cluster and an hypervisor configured
with ceph client to be able to do so. Another way to test could be to
verify that the generated XML for the VM volume is in the correct format
according to Libvirt RBD storage backend doc:
https://libvirt.org/storage.html#StorageBackendRBD

This functionality would be of great value to our deployments.

Thank you in advance,
Cynthia

2015-06-15 12:25 GMT+02:00 strzibny <notifications@github.com>:

lib/fog/libvirt/models/compute/templates/ceph.conf
```diff
@@ -0,0 +1,5 @@
+monitor=192.168.100.50,192.168.100.51,192.168.100.52
+port=6789
+libvirt_ceph_pool=foreman-images
+auth_username=foreman
+auth_uuid=8e1050ae-a53c-4280-bc48-8abe0eebafb1
```
plribeiro3000: Aren't those numbers specific to your enviroment? Maybe this file should be in `.gitignore`?

strzibny:
Support for the RBD is merged as https://github.com/fog/fog-libvirt/pull/56

PR #2, opened 2015-06-03
added support for generation of user-data iso. This is useful when os uses cloud-init for initial configuration.

witlessbird:


plribeiro3000:
Hey @witlessbird. Thank you for your patch!

Could you add some tests around it please?

Thanks!

witlessbird:
Is not using shindo for tests an option?

plribeiro3000:
The idea is to migrate all current tests to minitest and create new ones using minitest as well. Could you add test for this?

witlessbird:
minitest is perfect!

witlessbird:
- added support for mini-test tests and tests for user-data iso generation.
- will try to find time to see if volume-related tests can be written without excessive mocking.

lib/fog/libvirt/models/compute/server.rb
```diff
@@ -215,6 +219,35 @@ def vnc_port
...
+        def create_user_data_iso
+          generate_config_iso(user_data) do |iso|
+            vol = service.volumes.create(:name => cloud_init_volume_name, :capacity => "#{File.new(iso).size}b", :allocation => "0G")
+            vol.upload_image(iso)
+            @iso_file = cloud_init_volume_name
+          end
+        end
+
+        def cloud_init_volume_name
+          "#{name}-cloud-init.iso"
```
plribeiro3000: Does this need to be hardcoded?

I'm not that into `libvirt` itself but i wonder if everyone does use that same convention (`*-cloud-init.iso`).

plribeiro3000:
@witlessbird Could you take a look at the build error?

I believe you might have to pre install `libvirt` before the tests start to run.
`redcarpet` is failing to install as well.

lib/fog/libvirt/models/compute/server.rb
witlessbird: As opposed to user supplying the name for the user-data iso? Probably doable via some additional constructor parameter, but why would anyone be interested in something like that? 

No idea re: convention, but I figured this is pretty descriptive a name.

plribeiro3000: Hmm. this is just for creation process. So the name don't need to match anything related to the user usage. My bad.

I do agree the name is a good one. :+1:

witlessbird:
I can take a look at the build issues, but I think those should be handled in a separate PR, as it isn't specific to this PR.

plribeiro3000:
Sure. Once we get it up and running we cant get back here and see about it running. =)

witlessbird:
[test]

witlessbird:
rebased.

plribeiro3000:
Could you move your test namespace? Its on `test` folder when it should be `tests`.

witlessbird:
I didn't like the idea of two sets under one root, perhaps use ```minitests``` for it?

plribeiro3000:
Oh damm. got it. `minitest` would be good for now.

But don't skip `shindo` tests. Take a look [here](https://github.com/fog/fog-voxel/blob/master/Rakefile) for a quick idea of how to run both. Until we push everything to `minitest` we will have to run both.

witlessbird:
done and done. Default target now executes both shindo and mini-tests. ```minitest``` and ```test``` targets are specific to ```mini-test``` and ```shindo``` frameworks respectively.

plribeiro3000:
@witlessbird Could you take a look at the broken build?

witlessbird:
fixed build issues on 1.8.7.

plribeiro3000:
:+1:

plribeiro3000:
Thank you @witlessbird !

witlessbird:
thanks for the merge!

witlessbird:
@plribeiro3000: do you know when a new version of the gem is going to be released?

strzibny:
@witlessbird Not sure if there is a plan:). We can probably do a release with this feature.

witlessbird:
@strzibny: that'd be awesome. Lemme know if I can help with that.

plribeiro3000:
@witlessbird @strzibny One of the reasons to extract the providers is to have a complete independent release process. So we won't have to wait for `fog` to be released to be able to have this. =)

Do we need anything else here? If not i believe we could do a release. =)

strzibny:
@plribeiro3000 @witlessbird Okay, I will do a new release today.

witlessbird:
yay! Thank you muchly, @strzibny!

PR #3, opened 2015-06-08
Fixed build issues

witlessbird:
- libvirt-dev is now being installed during before_install phase
 - removed unused dev depedencies: redcarpet and thor

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: ruby
 
-sudo: false
+sudo: required
+
+before_install:
+ - sudo apt-get update -qq
+ - sudo apt-get install -y libvirt-dev
 
 script: bundle exec rake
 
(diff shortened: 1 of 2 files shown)


domcleal:
You might also need to move the two other JRuby tests to allowed_failures, or remove it altogether as ruby-libvirt just won't work with it.  mime-types also needs pinning to `< 2.4` in .gemspec when RUBY_VERSION is `< 1.9`.

witlessbird:
Looks like the latest mime-types that works with 1.8.7 is 1.25.1. Had to pin nokogiri, octokit, and rest-client too for builds on ruby 1.8.7.

witlessbird:
@plribeiro3000: ping?

plribeiro3000:
@witlessbird Sorry for the delay. Thanks you for taking the time to fix the build.
Could you add `jruby-head` as an allowed failure? We currently support it so we should not skip it at all.

Take a look here for more details: https://github.com/fog/fog/blob/master/.travis.yml.

witlessbird:
As @domcleal pointed out you cannot use fog-libvirt with jruby, as it relies on ruby-libvirt, which uses native extensions.

plribeiro3000:
Hmmm. Sorry for the confusion!

plribeiro3000:
Thanks @witlessbird and @domcleal !

witlessbird:
@plribeiro3000: cheers for the merge!

PR #9, opened 2015-11-12
make net-ssh require optional

geemus:


geemus:
Passing build here: https://travis-ci.org/fog/fog-libvirt/builds/90807047

PR #15, opened 2016-02-29
improve error handling when libvirt is being used by multiple processes at the same time

sford:
This is a followup to #14. I ran into same type of problems and had to fix several other areas to get `fog-libvirt` stable.

For more background, I had ~5 processes running that were bringing VMs up and down every couple minutes on one machine. The processes are often synced together, so like 5 VM creations would occur almost at the same time.

I have exercised these changes a lot... in the past day I have brought up/down a couple thousand VMs without too many problems.

### DIFF
--- a/lib/fog/libvirt/requests/compute/list_volumes.rb
+++ b/lib/fog/libvirt/requests/compute/list_volumes.rb
@@ -7,7 +7,13 @@ module Fog
           if filter.keys.empty?
             raw_volumes do |pool|
               pool.list_volumes.each do |volume_name|
-                data << volume_to_attributes(pool.lookup_volume_by_name(volume_name))
+                begin
+                  data << volume_to_attributes(pool.lookup_volume_by_name(volume_name))
+                rescue ::Libvirt::RetrieveError
+                  # Catch libvirt exceptions to avoid race conditions involving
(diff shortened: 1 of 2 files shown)


lib/fog/libvirt/requests/compute/list_volumes.rb
```diff
@@ -33,7 +39,7 @@ def volume_to_attributes(vol)
                 :allocation  => bytes_to_gb(vol.info.allocation),
                 :capacity    => bytes_to_gb(vol.info.capacity),
               }
-          rescue Libvirt::RetrieveError
+          rescue ::Libvirt::RetrieveError, ::Libvirt::Error
```
sford: I am not a Ruby expert, but I had problems with the change from #14. I was getting the following error:
```
..../gems/fog-libvirt-0.0.3/lib/fog/libvirt/requests/compute/list_volumes.rb:40:in `rescue in volume_to_attributes': uninitialized constant Fog::Compute::Libvirt::RetrieveError (NameError) (RuntimeError)
```

The actual exception being thrown is `Libvirt::RetreiveError`, so `Fog::Compute::` is not correct. I prepended the exception with `::` to get it working.

plribeiro3000:
LGTM.

PR #19, opened 2016-03-16
catch Libvirt exceptions in list_domains to protect against race conditions

sford:
This is a followup to #15. This fixes two more areas that are susceptible to Libvirt exceptions when libvirt is being used by multiple processes. For example, multiple instances of vagrant with the vagrant-libvirt plugin.

I have been running CI tests with this change for the past couple days with pretty good success.

### DIFF
--- a/lib/fog/libvirt/requests/compute/list_domains.rb
+++ b/lib/fog/libvirt/requests/compute/list_domains.rb
@@ -10,8 +10,25 @@ module Fog
           elsif filter.key?(:name)
             data << client.lookup_domain_by_name(filter[:name])
           else
-            client.list_defined_domains.each { |name| data << client.lookup_domain_by_name(name) } unless filter[:defined] == false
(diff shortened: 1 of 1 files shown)


sford:
Hey @plribeiro3000, any thoughts on this PR? Like #15 , this PR helps fix issues when using fog-libvirt in parallel.

plribeiro3000:
Maybe you could have a private method for the rescue part? I feel it vary hard to follow this code, specially with an unless in the end.

Maybe something like:

```ruby
def retrieve
  yield
rescue ::Libvirt::RetrieveError, ::Libvirt::Error
  nil
end
```

sford:
@plribeiro3000,

Thanks for the feedback! Took me a while to jump back on this, but I refactored the code to use `yield`. I think it is a lot cleaner.

plribeiro3000:
Yes. It is indeed. Thx for bearing with me on this. 👍

PR #20, opened 2016-03-23
Set iso_dir from image path in user-data ISO upload

domcleal:
When the default pool is configured to a location other than
/var/lib/libvirt/images, the iso_dir is now set to the uploaded location of the
user-data image rather than defaulting to default_iso_dir.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ sudo: required
 before_install:
  - sudo apt-get update -qq
  - sudo apt-get install -y libvirt-dev
+ - gem update bundler
 
 script: bundle exec rake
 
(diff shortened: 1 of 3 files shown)


dLobatog:
:+1: - it points to the right location with this fix

almereyda:
This seems important for thin provisioning. Is there any apparent reason why the Travis build fails, but the branch stays mergeable?

The diff doesn't seem to be too excessive.

domcleal:
The Travis CI build is failing due to a rubygems issue, which could probably be worked around by setting `sudo: false` or installing a newer version of rubygems in the script.

The merge check is just whether the patch merges cleanly.

plribeiro3000:
@domcleal Could you push that change into your PR so we can check if it fix the issue?

Thanks!

domcleal:
Sure, updated to install a newer version of Bundler (not rubygems, I misremembered) in the Travis CI script. Tests now pass.

plribeiro3000:
Thanks!

dLobatog:
How about a new patch release to include this? :smile:

plribeiro3000:
Just released as `0.2.0`. Sorry for the delay.

dLobatog:
Thanks!

PR #22, opened 2016-05-12
Allow the boot order to contain 'fd' (floppy).

### LINKED ISSUE #21: Allow floppy disks in boot order
At the moment we'll abort with an error if some domain exists and has `fd` in its boot order.

fred-a-kemp:
We shouldn't error out when some domain exists and has such a
configuration. Fixes:

closes #21

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -351,6 +351,10 @@ module Fog
           }
         end
 
+        def valid_boot_media
+          %w[cdrom fd hd network]
+        end
+
         def default_boot_order
           %w[hd cdrom network]
         end
@@ -358,7 +362,7 @@ module Fog
         def verify_boot_order order = []
           if order
             order.each do |b|
-              raise "invalid boot order, possible values are: hd, network and/or cdrom" unless default_boot_order.include?(b)
+              raise "invalid boot order, possible values are any combination of: #{valid_boot_media.join(', ')}" unless valid_boot_media.include?(b)
             end
           end
         end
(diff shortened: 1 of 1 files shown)


lib/fog/libvirt/models/compute/server.rb
```diff
@@ -351,14 +351,18 @@ def defaults
...
+        end
+
         def default_boot_order
           %w[hd cdrom network]
         end
 
         def verify_boot_order order = []
           if order
             order.each do |b|
-              raise "invalid boot order, possible values are: hd, network and/or cdrom" unless default_boot_order.include?(b)
```
plribeiro3000: Now that we are not using `default_boot_order` anymore, do we still need it?

fred-a-kemp: `:boot_order` is initialized to `default_boot_order` in the `defaults` method just a few lines up, so I guess it's still relevant. Admittedly, I have not looked at the full context of how any of this is used -- I only debugged this very locally/contextless.

plribeiro3000: Well, if it is not used anywhere else i believe we can get rid of this method and push the value to the DSL definition of `boot_order` instead.

What do you think?

fred-a-kemp: Agreed. Same applies to the `valid_boot_media()` method I introduced, so I removed that one as well. Please have a look at the new commit I just pushed.

fred-a-kemp: > Please have a look at the new commit I just pushed.

Ping?

plribeiro3000: Ouch. sorry for the delay.

plribeiro3000:
Thank you for your contribution!

PR #23, opened 2016-06-10
image_upload: Always return the number of bytes read

agx:
Currently volume uploads end up empty since ruby-libvirt returns the
value passed in from the block

    https://www.redhat.com/archives/libvir-list/2016-June/msg00678.html

Since this is zero libvirtd assumes the upload is finished before it
even started.

We can work around this by returning the string length. This will
continue to work with fixed ruby-libvirt since errors are indicated with
negative values.

### DIFF
--- a/lib/fog/libvirt/requests/compute/upload_volume.rb
+++ b/lib/fog/libvirt/requests/compute/upload_volume.rb
@@ -11,7 +11,7 @@ module Fog
           stream.sendall do |_opaque, n|
             begin
               r = image_file.read(n)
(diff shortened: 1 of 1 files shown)


plribeiro3000:
I'm not aware of this code so i can't confirm it is working.

Can you confirm this fixes the problem and that no other issues arise?

Thanks

agx:
In fact it should _unbreak_ uploading volumes and I'm not aware of any issues it can cause (we can revert should that happen), it's quiet isolated.

Issue #27, opened 2016-10-12
dhcp_leases.rb:8:in `dhcp_leases': undefined method `dhcp_leases' for #<Libvirt::Network:0x00000003ea12f0> (NoMethodError)


infernix:
`virNetworkGetDHCPLeases` isn't supported in libvirt < 1.2.6 so on distros carrying older libvirt, everything compiles but then using this call fails with above error.

This was introduced in PR #11 

I think a few things could be done:
1. Start depending on a specific libvirt version (depending on just ruby-libvirt doesn't seem to be enough)
2. Test if the method exists before calling it and then fail with a descriptive error if it doesn't
3. Support old and new behaviour (ugh)

Thoughts?

plribeiro3000:
Hey @infernix . Thanks for the feedback.

I believe option 2 would be the best option as we don't have an active maintainer on this repo.

Would you be able to do a patch and submit a PR for this change?

infernix:
@plribeiro3000 the little time I have for coding I'm spending on vagrant-libvirt, so can't make any promises but will take a look.

plribeiro3000:
Awesome. Thank you! :+1:

Em seg, 17 de out de 2016 08:12, Gerben Meijer <notifications@github.com>
escreveu:

PR #29, opened 2016-11-09
Support older libvirt address retrieval

electrofelix:
Libvirt releases before 1.2.8 do not fully support the
virNetworkGetDHCPLeases used by dhcp_leases, so provide a compatibility
check to determine the version of the library in order to determine
which method to use to retrieve the mac address and then replace the
current addresses method reference with this.

Currently uses calling 'virsh verison' and regex comparison of the response to
determine whether to use the previous approach which has been restored
or the more recent method that calls dhcp_leases.

Fixes: #27

### DIFF
--- a/lib/fog/libvirt/compute.rb
+++ b/lib/fog/libvirt/compute.rb
@@ -45,6 +45,7 @@ module Fog
       request :destroy_interface
       request :get_node_info
       request :update_display
+      request :libversion
 
       module Shared
         include Fog::Compute::LibvirtUtil
@@ -57,11 +58,11 @@ module Fog
           require 'libvirt'
         end
 
+        private
+
         def client
           return @client if defined?(@client)
         end
-
-        private
 
         #read mocks xml
         def read_xml(file_name)
--- a/lib/fog/libvirt/compute.rb
+++ b/lib/fog/libvirt/compute.rb
@@ -57,11 +57,11 @@ module Fog
           require 'libvirt'
         end
 
-        private
-
         def client
           return @client if defined?(@client)
         end
+
+        private
 
         #read mocks xml
         def read_xml(file_name)
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -278,7 +278,7 @@ module Fog
 
           # if ruby-libvirt not compiled with support, or remote library is
           # too old (must be newer than 1.2.8), then use old fallback
-          if not has_dhcp_leases or service.client.libversion < 1002008
+          if not has_dhcp_leases or service.libversion() < 1002008
             addresses_method = self.method(:addresses_ip_command)
           end
 
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -258,11 +258,134 @@ module Fog
         private
         attr_accessor :volumes_path
 
-        # This retrieves the ip address of the mac address
+        # This tests the library version before redefining the address
+        # method for this instance to use a method compatible with
+        # earlier libvirt libraries, or uses the dhcp method from more
+        # recent releases.
+        def addresses(service_arg=service, options={})
+          addresses_method = self.method(:addresses_dhcp)
+          # check if ruby-libvirt was compiled against a new enough version
+          # that can use dhcp_leases, as otherwise it will not provide the
+          # method dhcp_leases on any of the network objects.
+          has_dhcp_leases = true
+          begin
+            service.networks.first.dhcp_leases(self.mac)
+          rescue NoMethodError
+            has_dhcp_leases = false
+          rescue
+            # assume some other odd exception.
+          end
+
+          # if ruby-libvirt not compiled with support, or remote library is
+          # too old (must be newer than 1.2.8), then use old fallback
+          if not has_dhcp_leases or service.client.libversion < 1002008
+            addresses_method = self.method(:addresses_ip_command)
+          end
+
+          # replace current definition for this instance with correct one for
+          # detected libvirt to perform check once for connection
+          (class << self; self; end).class_eval do
+            define_method(:addresses, addresses_method)
+          end
+          addresses(service_arg, options)
+        end
+
+        # This retrieves the ip address of the mac address using ip_command
         # It returns an array of public and private ip addresses
         # Currently only one ip address is returned, but in the future this could be multiple
         # if the server has multiple network interface
-        def addresses(service_arg=service, options={})
+        def addresses_ip_command(service_arg=service, options={})
+          mac=self.mac
+
+          # Aug 24 17:34:41 juno arpwatch: new station 10.247.4.137 52:54:00:88:5a:0a eth0.4
+          # Aug 24 17:37:19 juno arpwatch: changed ethernet address 10.247.4.137 52:54:00:27:33:00 (52:54:00:88:5a:0a) eth0.4
+          # Check if another ip_command string was provided
+          ip_command_global=service_arg.ip_command.nil? ? 'grep $mac /var/log/arpwatch.log|sed -e "s/new station//"|sed -e "s/changed ethernet address//g" |sed -e "s/reused old ethernet //" |tail -1 |cut -d ":" -f 4-| cut -d " " -f 3' : service_arg.ip_command
+          ip_command_local=options[:ip_command].nil? ? ip_command_global : options[:ip_command]
+
+          ip_command="mac=#{mac}; server_name=#{name}; "+ip_command_local
+
+          ip_address=nil
+
+          if service_arg.uri.ssh_enabled?
+
+            # Retrieve the parts we need from the service to setup our ssh options
+            user=service_arg.uri.user #could be nil
+            host=service_arg.uri.host
+            keyfile=service_arg.uri.keyfile
+            port=service_arg.uri.port
+
+            # Setup the options
+            ssh_options={}
+            ssh_options[:keys]=[ keyfile ] unless keyfile.nil?
+            ssh_options[:port]=port unless keyfile.nil?
+            ssh_options[:paranoid]=true if service_arg.uri.no_verify?
+
+            begin
+              result=Fog::SSH.new(host, user, ssh_options).run(ip_command)
+            rescue Errno::ECONNREFUSED
+              raise Fog::Errors::Error.new("Connection was refused to host #{host} to retrieve the ip_address for #{mac}")
+            rescue Net::SSH::AuthenticationFailed
+              raise Fog::Errors::Error.new("Error authenticating over ssh to host #{host} and user #{user}")
+            end
+
+            # Check for a clean exit code
+            if result.first.status == 0
+              ip_address=result.first.stdout.strip
+            else
+              # We got a failure executing the command
+              raise Fog::Errors::Error.new("The command #{ip_command} failed to execute with a clean exit code")
+            end
+
+          else
+            # It's not ssh enabled, so we assume it is
+            if service_arg.uri.transport=="tls"
+              raise Fog::Errors::Error.new("TlS remote transport is not currently supported, only ssh")
+            end
+
+            # Execute the ip_command locally
+            # Initialize empty ip_address string
+            ip_address=""
+
+            IO.popen("#{ip_command}") do |p|
+              p.each_line do |l|
+                ip_address+=l
+              end
+              status=Process.waitpid2(p.pid)[1].exitstatus
+              if status!=0
+                raise Fog::Errors::Error.new("The command #{ip_command} failed to execute with a clean exit code")
+              end
+            end
+
+            #Strip any new lines from the string
+            ip_address=ip_address.chomp
+          end
+
+          # The Ip-address command has been run either local or remote now
+
+          if ip_address==""
+            #The grep didn't find an ip address result"
+            ip_address=nil
+          else
+            # To be sure that the command didn't return another random string
+            # We check if the result is an actual ip-address
+            # otherwise we return nil
+            unless ip_address=~/^(\d{1,3}\.){3}\d{1,3}$/
+              raise Fog::Errors::Error.new(
+                        "The result of #{ip_command} does not have valid ip-address format\n"+
+                            "Result was: #{ip_address}\n"
+                    )
+            end
+          end
+
+          return { :public => [ip_address], :private => [ip_address]}
+        end
+
+        # This retrieves the ip address of the mac address using dhcp_leases
+        # It returns an array of public and private ip addresses
+        # Currently only one ip address is returned, but in the future this could be multiple
+        # if the server has multiple network interface
+        def addresses_dhcp(service_arg=service, options={})
           mac=self.mac
 
           ip_address = nil
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -290,6 +290,56 @@ module Fog
           addresses(service_arg, options)
         end
 
+        def ssh_ip_command(command, uri = {})
+
+          # Retrieve the parts we need from the service to setup our ssh options
+          user=uri.user #could be nil
+          host=uri.host
+          keyfile=uri.keyfile
+          port=uri.port
+
+          # Setup the options
+          ssh_options={}
+          ssh_options[:keys]=[ keyfile ] unless keyfile.nil?
+          ssh_options[:port]=port unless keyfile.nil?
+          ssh_options[:paranoid]=true if uri.no_verify?
+
+          begin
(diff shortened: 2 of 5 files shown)


electrofelix:
So while this works, I've been struggling to adjust the tests to allow for checking the version. I'm also wondering if there is a better way to check the libvirt library version of the connection being used as opposed to assuming a local version check?

plribeiro3000:
Hmm. This is quite complex to review code only.

Since i don't have a `libvirt` environment to test this i believe we will have to rely on @strzibny to check this out. Sorry about that.

lib/fog/libvirt/models/compute/server.rb
```diff
@@ -257,11 +257,132 @@ def cloud_init_volume_name
...
+          end
+          addresses(service_arg, options)
+        end
+
+        # This retrieves the ip address of the mac address using ip_command
         # It returns an array of public and private ip addresses
         # Currently only one ip address is returned, but in the future this could be multiple
         # if the server has multiple network interface
-        def addresses(service_arg=service, options={})
+        def addresses_ip_command(service_arg=service, options={})
```
electrofelix: Note that this is a restoration of the original code that was replaced when switching to dhcp_leases.

lib/fog/libvirt/models/compute/server.rb
```diff
@@ -257,11 +257,128 @@ def cloud_init_volume_name
...
+        # method for this instance to use a method compatible with
+        # earlier libvirt libraries, or uses the dhcp method from more
+        # recent releases.
+        def addresses(service_arg=service, options={})
+          addresses_method = self.method(:addresses_dhcp)
+          # check if ruby-libvirt was compiled against a new enough version
+          # that can use dhcp_leases, as otherwise it will not provide the
+          # method dhcp_leases on any of the network objects.
+          network = service.client.lookup_network_by_uuid(
+            service.networks.first.uuid)
```
electrofelix: I suspect that part of this and the call below to `service.client.libversion` both need to be moved to the requests file in order to allow existing tests to pass and follow the current coding structure where the code that actually talks to libvirt is all in requests?

electrofelix:
@plribeiro3000 @strzibny looks like I've got all the test issues ironed out, and also looks like I've a better way to detect firstly if ruby-libvirt is compiled with support for dhcp_leases, and secondly whether the connection is talking to a libvirt API that is new enough to respond to such a call.

Only question now remains, are you happy to restore the old approach to maintain compatibility with libvirt older than 1.2.8?

lib/fog/libvirt/compute.rb
```diff
@@ -57,12 +57,16 @@ def initialize(options={})
           require 'libvirt'
         end
 
-        private
-
         def client
-          return @client if defined?(@client)
+          @client ||= Class.new do
+            def libversion
+              1002009
```
plribeiro3000: Why do you need this to be hardcoded?

lib/fog/libvirt/models/compute/server.rb
```diff
@@ -257,11 +257,134 @@ def cloud_init_volume_name
...
         # Currently only one ip address is returned, but in the future this could be multiple
         # if the server has multiple network interface
-        def addresses(service_arg=service, options={})
+        def addresses_ip_command(service_arg=service, options={})
+          mac=self.mac
+
+          # Aug 24 17:34:41 juno arpwatch: new station 10.247.4.137 52:54:00:88:5a:0a eth0.4
+          # Aug 24 17:37:19 juno arpwatch: changed ethernet address 10.247.4.137 52:54:00:27:33:00 (52:54:00:88:5a:0a) eth0.4
+          # Check if another ip_command string was provided
+          ip_command_global=service_arg.ip_command.nil? ? 'grep $mac /var/log/arpwatch.log|sed -e "s/new station//"|sed -e "s/changed ethernet address//g" |sed -e "s/reused old ethernet //" |tail -1 |cut -d ":" -f 4-| cut -d " " -f 3' : service_arg.ip_command
```
plribeiro3000: Maybe you should just return nil in case `ip_command` is nil?

I wonder where all use cases will have the same path and all logs are the same way. Maybe the case where it is empty should be left to the user of the library to decide?

lib/fog/libvirt/models/compute/server.rb
```diff
@@ -257,11 +257,134 @@ def cloud_init_volume_name
...
+            # otherwise we return nil
+            unless ip_address=~/^(\d{1,3}\.){3}\d{1,3}$/
+              raise Fog::Errors::Error.new(
+                        "The result of #{ip_command} does not have valid ip-address format\n"+
+                            "Result was: #{ip_address}\n"
+                    )
+            end
+          end
+
+          return { :public => [ip_address], :private => [ip_address]}
```
plribeiro3000: There is plenty of logic in this method that should go into a request instead.
There should be only business logic here. All the `ssh` calls and `IO` should be on a request.

lib/fog/libvirt/compute.rb
electrofelix: A test was added to `tests/libvirt/models/compute/server_tests.rb` with the switch to dhcp_leases:
https://github.com/fog/fog-libvirt/blob/558c05d9d6ca6e764e03aa450b57b2728a7f3404/tests/libvirt/models/compute/server_tests.rb#L23-L25

I couldn't find a better way to help that only go through the dhcp_leases code by default. Happy to switch if there is another way, just not familiar enough with shindo testing framework to know how to solve this better.

lib/fog/libvirt/models/compute/server.rb
electrofelix: This all was working for most users before this code was originally removed, and when needed it was being overridden, so it seems like this is just providing a default and trusting people to set the appropriate command to use instead as needed.

If going to re-enable support of older libvirt versions, it seems like it would be less likely to cause problems for anyone pinned to an earlier version of fog-libvirt if we avoid changing the default behaviour of the old code, as that would allow them to move forward by just unpinning instead of needing to copy the command into their application as well.

lib/fog/libvirt/models/compute/server.rb
electrofelix: I've been reluctant to alter this code beyond restoring it from the commit that removed it and providing a entry point that allows it to work with older libvirt libraries until such point that there is no point in supporting them. Perhaps it would be better to say whether this restoration will be acceptable in principal and then we can work with additional commits to move pieces of this to requests.

Does that seem sensible?

electrofelix:
Anyone able to comment on whether this will be fundamentally acceptable before I start working on changing the old restored code to use the request model?

lib/fog/libvirt/compute.rb
plribeiro3000: The links you pasted is testing the method `public_ip_address`. Is that the root of this change above?

plribeiro3000:
@electrofelix Can you point me to the code that introduced the problem you are having?
I dont use libvirt anymore so i need to triple check everything before i merge.

electrofelix:
@plribeiro3000 It's from a number of users running vagrant-libvirt with an older libvirt that doesn't support the newer call `dhcp_leases` as described in #27. Unfortunately some of these distributions are in maintenance mode such as RHEL and Ubuntu 14.04. I'll try to put together a Vagrantfile that can use virtualbox to demonstrate the problem when running ubuntu 14.04 fully patched if that would help?

That should at least allow others to see the issue

plribeiro3000:
No. thats ok. We don't need the extra effort. I just wanted to make sure the reason behind this is documented somewhere. Since there is an issue for it already we don't need anything else.

Could you just answer my question [here](https://github.com/fog/fog-libvirt/pull/29/files#r106888137)? I'm really concerned about that hardcoded number.

lib/fog/libvirt/compute.rb
electrofelix: Yes, I have to admit I struggled a bit trying to understand how best to work with Shindo tests, have you any suggestions on how to mock/fake that the test there should be testing the code as though it's using a libvirt library version greater than 1.2.8 (1002009 corresponds to 1.2.9)?

plribeiro3000: Dunno but i feel this should be done inside tests only so it wont change the behaviour of a real execution.
Maybe you can push this part of the code to a initializer that runs before a specific tests for libvirt `1.2.9`?

electrofelix: Would I be wrong in thinking in terms of Shindo tests this is the Mock class to be used by the tests?

I'll see if I can do anything else with it, maybe try setting it from within the test code itself.

plribeiro3000: Yes. you would be wrong. =)

The Mock class is not for test purposes directly but rather a feature fog has so you can try your code without actually accessing the provider api (It stores all data inside objects so you can get the result on a later call).

Your proposed solution (Set it from within the test code itself) would fit better. =)

infernix:
@electrofelix @plribeiro3000 thanks for the effort here! But can this be merged at some point? If not I'm afraid we will have to deprecate vagrant-libvirt support for older distros if not :(

electrofelix:
@infernix thanks for the reminder, I've hacked around what I think is the last issue, mainly it's been the difficulty in understanding how to mock the libversion for one of the shindo tests that was causing me the most problems.

If the updated fix makes sense I'll provide some additional changes to help tidy up the restored code based on some of the comments until such time as we no longer need support for the earlier versions of libvirt.

@plribeiro3000 is what I've put in place to patch in a return value for the libversion during testing make sense? or is there a better way to add that in? I had to allow the client attribute to be public, but seeing as it appears to be public in the real object, I'm hoping that's not a problem.

plribeiro3000:
@electrofelix @infernix Can you guys confirm this work as intended?

I'm a bit worried about this code. The methods are too big and not in the place they belong (requests). Since i'm not working with `libvirt` for quite a while i'm not able to tell if this change makes sense at all (So much going on here).

If you guys say that this does work, i will have only one question. Does this code make the `client` method public just for testing purposes?

electrofelix:
@infernix thanks for the reminder, I've hacked around what I think is the last issue, mainly it's been the difficulty in understanding how to mock the libversion for one of the shindo tests that was causing me the most problems.

If the updated fix makes sense I'll provide some additional changes to help tidy up the restored code based on some of the comments until such time as we no longer need support for the earlier versions of libvirt.

@plribeiro3000 is what I've put in place to patch in a return value for the libversion during testing make sense? or is there a better way to add that in? I had to allow the client attribute to be public, but seeing as it appears to be public in the real object, I'm hoping that's not a problem.

infernix:
@electrofelix did you intend to write the exact same comment twice? :)

electrofelix:
Nope, I'll delete it, was still in the GH comment field, so I assumed I didn't send it the first time.

electrofelix:
@plribeiro3000 Inside the existing `Fog::Compute::Libvirt::Real` class the client attribute is publically available https://github.com/fog/fog-libvirt/pull/29/files#diff-d0d400fc75bff5967bdd315151c45251L75

So wasn't too sure why the `Fog::Compute::Libvirt::Mock` code made it private?

I'm happy to add another commit on top of this to refactor the code into using the request approach before it's landed, just wanted to check that I've done to fix up the testing side makes sense. Already had a look at that, don't think it will be too difficult.

infernix:
@plribeiro3000 could you provide some feedback? Thanks!

plribeiro3000:
@infernix Sorry i took too long to review this.

Gotta say i'm not really comfortable with the changes added here. I do know that the method `addresses` was a mess before but i don't feel we should put it back as a mass again just because.

Maybe we can try to break `addresses_ip_command` into smaller methods?
I've saw at least 2 parts of the code that could become smaller methods that will make this easier to maintain.

Is that something we can work on before merging this in?

electrofelix:
@plribeiro3000 finally got around to sorting out some of the refactoring, split most of address_ip_command into two helpers, one local and one remote. Not sure if this is what you meant?

Also moved the libversion call to a requests so I could remove the changes to the shindo tests and switched the Mock client attribute back to being private.

Looks like there might be an issue with library compatibility around the fog-core/fog-json dependencies that is breaking for some of the test matrices.

plribeiro3000:
@electrofelix I believe the split in helpers was a good move. The private move on `Mock` client attribute was also a good move. 👍 for those refactorings. Maybe we can improve it a little bit more?

It seems that the final piece of code in `#addresses_ip_comand` does quite a few things:

* It initializes as `nil`
* If ssh is enabled it does get it by ssh
* raises an exception if is a `tls` encrypted connection
* get the ip locally
* raise exception if ip is invalid

I believe we can break this flow a little bit like:

* Drop the `nil` initialization
* Check for `TLS` connection from within the method that gets the ip locally
* Raise exception if ip is invalid from within `#addresses_ip_comand`

This would make `#addresses_ip_comand` a simple `if else` to get the ip and check if the result is valid

And you can simplify `addresses` to be something like:

```ruby
def addresses(service_arg=service, options={})
  service.networks.first.dhcp_leases(self.mac)
  return addresses_ip_command if service.libversion() < 1002008
  addresses_dhcp(service_arg, options)
rescue NoMethodError
  addresses_ip_command(service_arg, options)
end
```

You could also improve `#ssh_ip_command` by rescueing the exceptions at the end like the example above.

This modifications would make the code simplier and better to maintain.

electrofelix:
ok, so a bit more tidy up to be done.

The existing code I added for `addresses` takes care of replacing the initial method with the correct target method so that all future calls will go direct instead of needing to check on each invocation of this call on the same connection. I was thinking that for any remote connections, that trying to call dhcp_leases each time to determine it doesn't exist, before verifying the library version might be somewhat expensive in latency.

I'm happy enough not to worry about this myself and simplify the code as I don't use it for remote connections, can leave it for someone to report an issue, or maybe just create a PR subsequently and we can discuss the performance versus maintenance pros/cons separately?

On the 'TLS' connection test, I was assuming this would require another method of the form tls_ip_command, which would mean calling it from the address_ip_command and removing the code from the local_ip_command to error out? So it did seem a little odd to move it there in the first place, but maybe I should move it to a separate elseif part of the if/else block so it's not part of the 'else' statement? I'll put that up and we can discuss further.

I also spotted you had a few comments about moving the SSH/IO calls out to requests as well from an earlier review, and wondering how to do that, as most of the Requests appear to be around use of the libvirt client to perform remote requests. Any examples elsewhere that would have a good guideline on what to do?

plribeiro3000:
Ok. This is open for quite a while now. @electrofelix do you feel strong about merging this in?
We can discuss the other points in other PR's.

Now that you mentioned on the TLS and remote connections i think the code is ok for now. It would be better to bring it in and discuss its improvement in another PR.

Regarding using requests i don't have any examples. `fog` is so big that each provider has its own way to do it nowadays but i believe the best approach would be to leave the models as thin layers of logic and the requests would be anything that touches a service outside `fog`.
This would make all requests simple enough so they would not need that much test. And all the logic like parsing data, generating data to send to the provider, change a state and so on can be easily tested without depending on any http requests or any sort of problematic dependency.

infernix:
Merging this in fixes libvirt for older distros such as debian 8, centos 6, ubuntu 12.xx, 14.xx (iirc). It does become less relevant over time but some of those are LTS so would be good to have it just work[tm] on them.

plribeiro3000:
Coll. Bring it in!

infernix:
So this is working great on older distros but with a minor patch:

```diff
diff --git a/lib/fog/libvirt/models/compute/server.rb b/lib/fog/libvirt/models/compute/server.rb
index cecaf5b..7fd663d 100644
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -321,7 +321,7 @@ module Fog
           end
         end
 
-        def local_ip_command(command)
+        def local_ip_command(ip_command)
           # Execute the ip_command locally
           # Initialize empty ip_address string
           ip_address=""
```

If not patched, it's throwing:
```
lib/fog/libvirt/models/compute/server.rb:329:in `local_ip_command': undefined local variable or method `ip_command' for #<Fog::Compute::Libvirt::Server:0x0000000002475048> (NameError)
    from /root/.vagrant.d/gems/2.4.3/gems/fog-libvirt-0.4.2/lib/fog/libvirt/models/compute/server.rb:367:in `addresses_ip_command'
    from /root/.vagrant.d/gems/2.4.3/gems/fog-libvirt-0.4.2/lib/fog/libvirt/models/compute/server.rb:290:in `addresses'
    from /root/.vagrant.d/gems/2.4.3/gems/vagrant-libvirt-0.0.43/lib/vagrant-libvirt/action/wait_till_up.rb:43:in `block (3 levels) in call'
...
```


@electrofelix can you confirm this is a typo?

electrofelix:
@infernix it is, and it seems the tests I added are not testing the ssh_ip_command and local_ip_command methods sufficiently, otherwise they should have both thrown an error on this, working on a fix

PR #30, opened 2016-11-21
Allow update_display on VMs with non-auto keymap

tbrisker:
If the vnc keymap has been changed from auto to, for example, en-us,
calling update_display will fail with a "Call to
virDomainUpdateDeviceFlags failed: internal error cannot change keymap
setting on vnc graphics" error. This commit changes it so that if no
keymap has been passed explicitly in the options we will use the keymap
that is currently used.

### DIFF
--- a/lib/fog/libvirt/requests/compute/update_display.rb
+++ b/lib/fog/libvirt/requests/compute/update_display.rb
@@ -4,17 +4,21 @@ module Fog
       class Real
         def update_display(options = { })
           raise ArgumentError, "uuid is a required parameter" unless options.key? :uuid
+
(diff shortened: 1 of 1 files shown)


plribeiro3000:
@strzibny Could you please take a look at this PR?
I'm afraid i can't just look at the code to merge this.

mmoll:
What's the status here?

BTW, there are quite some people from @theforeman using fog-libvirt actively, if new maintainers need to be found...

strzibny:
This looks good to me, I will merge. @mmoll it would be awesome if some folks from Foreman wants to co-maintain :) (anyone can shoot me an e-mail).

mmoll:
☝️ @lzap @tbrisker ☝️

PR #31, opened 2016-11-23
fix error raising for cloud-init genisoimage

mmoll:
Also adjust the error message so it's easier to find the cause. I didn't have genisoimage installed on the host in question...


plribeiro3000:
Thanks @mmoll .

Could you use the standard format instead?

```ruby
raise Fog::Errors::Error, "Couldn't generate cloud-init iso disk with genisoimage."
```

mmoll:
@plribeiro3000 I can do that, but a quick "git grep" over fog-libvirt shows zero occourences of that format, maybe the others should get changed then, too?

plribeiro3000:
Hmm. Yeah, they can be changed but we can leave this to another PR then.

👍

plribeiro3000:
Thanks for your patience. =)

PR #32, opened 2016-11-24
Fix breaking CI on older versions

tbrisker:
Pin some dependencies so that travis can run on older ruby versions.


plribeiro3000:
Thanks @tbrisker !

Issue #18, opened 2016-03-11
Support for cpu mode


rski:
Hi,

I'd like to be able to set the cpu mode on my VMs on creation, but fog-libvirt doesn't seem to currently support this. This would require adding a line like that can render to `<cpu mode='something'/>` to `fog-libvirt/lib/fog/libvirt/models/compute/templates/server.xml.erb`, along with other things.

Is this the right place to look for a feature like this? If yes, would it be ok if I hacked on this and submitted a PR to add this functionality?

PR #38, opened 2017-04-02
implement cpu model and mode selection

mmoll:
closes GH-18

### DIFF
--- a/lib/fog/libvirt/models/compute/templates/server.xml.erb
+++ b/lib/fog/libvirt/models/compute/templates/server.xml.erb
@@ -13,6 +13,23 @@
     <apic/>
     <pae/>
   </features>
+<% if !cpu.empty? -%>
+<% if cpu[:mode] -%>
+  <cpu mode='<%= cpu[:mode] %>'>
+<% else -%>
+  <cpu>
+<% end -%>
+<%
+  fallback = 'allow'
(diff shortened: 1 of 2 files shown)


mmoll:
The most common use case is to set `:cpu=>{:mode=>"host-model"}`.

I tried to keep this extendable if further cpu parameters will get added, but would state the current ERB approach gave not the best result.

mmoll:
anybody? :)

lib/fog/libvirt/models/compute/templates/server.xml.erb
```diff
@@ -13,6 +13,23 @@
     <apic/>
     <pae/>
   </features>
+<% if !cpu.empty? -%>
+<% if cpu[:mode] -%>
```
plribeiro3000: Do you need to check for both emptiness and the `mode` key?

mmoll: The situation is a bit messy (https://libvirt.org/formatdomain.html#elementsCPU) and for the sake for future-proveness (in theory more parameters could be supported) and setting default values I did it that way.

plribeiro3000: Hmmm. I guess you are right here. =)

PR #39, opened 2017-04-06
Skip over networks that no longer exist

omahn:
If a user is running multiple libvirt clients concurrently it's possible to get into a situation where `list_networks.rb` will attempt to lookup a network that has been removed by another client. We hit this scenario ourselves today as we are rapidly creating and destroying both domains and virtual networks for testing purposes. This PR will simply skip over any networks that no longer exist but will still raise an exception for any other errors returned when attempting to lookup networks.

### DIFF
--- a/lib/fog/libvirt/requests/compute/list_networks.rb
+++ b/lib/fog/libvirt/requests/compute/list_networks.rb
@@ -6,7 +6,15 @@ module Fog
           data=[]
           if filter.keys.empty?
             (client.list_networks + client.list_defined_networks).each do |network_name|
-              data << network_to_attributes(client.lookup_network_by_name(network_name))
+              begin
+                data << network_to_attributes(client.lookup_network_by_name(network_name))
+              rescue => error
+                if error.message =~ /Network not found/
+                  next # Network no longer exists, skip over and carry on.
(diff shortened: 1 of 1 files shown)


lib/fog/libvirt/requests/compute/list_networks.rb
```diff
@@ -6,7 +6,15 @@ def list_networks(filter = { })
           data=[]
           if filter.keys.empty?
             (client.list_networks + client.list_defined_networks).each do |network_name|
-              data << network_to_attributes(client.lookup_network_by_name(network_name))
+              begin
+                data << network_to_attributes(client.lookup_network_by_name(network_name))
+              rescue => error
+                if error.message =~ /Network not found/
```
plribeiro3000: It raises an Generic exception for all errors?

plribeiro3000:
Hey @omahn. sorry for the big delay here.

And again, i'm sorry for not being as helpul as i should but i'm not fully aware of this project's code so i need your help for some details (like the generic exception question).

strzibny:
Shouldn't `list_networks` method just accept this as an option? (`skip_missing_networks:` for example)

Issue #42, opened 2017-07-04
Operation not supported: cannot change keymap setting on vnc graphics


santachago:
Looks like libvirt doesnt support to set keymap when using vnc.

``` | Libvirt::Error: Call to virDomainUpdateDeviceFlags failed: Operation not supported: cannot change keymap setting on vnc graphics
 | Libvirt::Error: Call to virDomainUpdateDeviceFlags failed: Operation not supported: cannot change keymap setting on vnc graphics
 | /usr/share/foreman/vendor/ruby/2.3.0/gems/fog-libvirt-0.4.0/lib/fog/libvirt/requests/compute/update_display.rb:21:in `update_device'
 | /usr/share/foreman/vendor/ruby/2.3.0/gems/fog-libvirt-0.4.0/lib/fog/libvirt/requests/compute/update_display.rb:21:in `update_display'
 | /usr/share/foreman/vendor/ruby/2.3.0/gems/fog-libvirt-0.4.0/lib/fog/libvirt/models/compute/server.rb:218:in `update_display'
```

mmoll:
@santachago could you check if reverting #30 fixes that issue?

mmoll:
@tbrisker looking at #30 I don't get the whole current situation, as that was also about vnc

tbrisker:
@mmoll IIRC #30 in my case was caused by having a non-auto setting for the keymap, which was failing because fog-libvirt wasn't passing a keymap and libvirt was defaulting to auto (which was different then the existing setting and thus failed to change it). That being said, I think there may be a better solution then this. Sadly I don't have the setup I had previously to try and fix it at the moment.

santachago:
If i remove line number 16 -> display[:keymap] = options[:keymap] || xml_elements(domain.xml_desc, "graphics", "keymap")

vms with no keymap defined on the config work fine. If a vm has a value defined it fails with the same error.

mmoll:
possible fix in #43

plribeiro3000:
@santachago @tbrisker @mmoll Could you guys please test if master fixed the issue please?

mmoll:
I can only say it fixed it for me (which is expected as I wrote the patch ;)).

mmoll:
@santachago any update here? ☝️

plribeiro3000:
@mmoll Thanks for the feeback. 👍

hesperaux:
Hello all,

[...]

PR #43, opened 2017-08-05
don't try to set a keymap when there's none

mmoll:
closes GH-42


mmoll:
As a bonus point fix that, at least on my setup, `xml_elements(domain.xml_desc, "graphics", "keymap")` is returning something like `["en-us"]`.

plribeiro3000:
Sorry for the huge delay to review your changes.

Thank you for your efforts on this!

Issue #6, opened 2015-08-02
hugepages support


plribeiro3000:
Hi,

Would it be possible to optionally enable `hugepages` support for the libvirt driver? This could default to `false`, but having an option to enable it when creating the VM would be great.

When enabled, the `server.xml` template would have:

``` xml
<domain>
  <memoryBacking>
    <hugepages/>
  </memoryBacking>
  ...
</domain>
```

Enabling hugepages support on a Linux system requires a few steps, but I'm guessing documenting them would be outside of fog's scope.

Thank you.

Original issue opened by @eolamey at fog/fog#3342

cc/ @eolamey @geemus

eolamey:
Hi,

I am no longer needing this, should this issue be closed?

plribeiro3000:
Hey @eolamey. Thank you for the feedback but i believe those issues could live here a little longer. Now that we have a small repository it should be easier to add those features. =)

Thanks!

PR #45, opened 2017-09-08
libvirt template: Add hugepages support

rfrht:
Hi there guys;

While checking issue #6, I thought that this PR would fix it.
Notice that the code is untested: Used commit bdccac71 as a template.

If the attribute 'hugepages' is set to any value, then define the guest as
having its memory backed by a hugepages pool.
Resolves: fog/fog-libvirt#6

### DIFF
--- a/lib/fog/libvirt/models/compute/templates/server.xml.erb
+++ b/lib/fog/libvirt/models/compute/templates/server.xml.erb
@@ -1,6 +1,11 @@
 <domain type='<%= domain_type %>'>
   <name><%= name %></name>
   <memory><%= memory_size %></memory>
+<% if !hugepages.empty? -%>
+  <memoryBacking>
+    <hugepages/>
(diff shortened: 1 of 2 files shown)


lib/fog/libvirt/models/compute/templates/server.xml.erb
```diff
@@ -1,6 +1,11 @@
 <domain type='<%= domain_type %>'>
   <name><%= name %></name>
   <memory><%= memory_size %></memory>
+<% if !hugepages.empty? -%>
```
mmoll: I think that should rather be a boolean and then handled like `if hugepages`

strzibny: I agree with @mmoll. Can you change this?

strzibny:
I think we can close this in favour of #65

PR #47, opened 2017-10-19
Drop Ruby<2.0 support

tbrisker:
Users requiring ruby 1.9 support should use fog gem. No need to support
it here.


plribeiro3000:
@tbrisker Could you explicit require `ruby > 2.0` in the gemspec as well please?
That should help bundler install the correct version of `fog-libvirt`.

tbrisker:
@plribeiro3000 thanks, updated with minimum ruby version.

plribeiro3000:
Thanks!

PR #52, opened 2018-03-22
Fix tests for latest minitest

electrofelix:
Minitest has changed behaviour in new releases in how to mock and stub
around blocks.


electrofelix:
Seems my version of minitest gem that I tested with locally didn't match what Travis was using :-(

plribeiro3000:
Thats ok. =)

PR #57, opened 2018-07-14
Add qemu-guest-agent channel.

### LINKED ISSUE #55: qemu-guest-agent support
`server.xml.erb` is not very configurable right now. 
Adding the lines:
```
    <channel type='unix'>
      <target type='virtio' name='org.qemu.guest_agent.0'/>
    </channel>
```
allows `qemu-guest-agent` connectivity to the VM, which is crucial for `fsfreeze`, `trim` and other commands. 
If this is present on VM creation, e.g. `RHEL 7.5` / `CentOS 7.5` systems install the guest agent with anaconda, and set up the Guest Agent Service automatically.

olifre:
Fixes #55.

This adds a `qemu-guest-agent` channel. 
It allows to instruct the VM from the hypervisor e.g. to freeze filesystems for snapshotting. 
With RHEL 7 / CentOS 7, if the agent channel is present, anaconda installs and activates the guest agent automatically. 

People might want to configure this (on / off), but I would vote to have this on by default

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -29,6 +29,7 @@ module Fog
         attribute :display
         attribute :cpu
         attribute :hugepages
+        attribute :guest_agent
 
         attribute :state
 
@@ -480,6 +481,7 @@ module Fog
             :display                => default_display,
             :cpu                    => {},
             :hugepages              => false,
+            :guest_agent            => true,
           }
         end
 
--- a/lib/fog/libvirt/models/compute/templates/server.xml.erb
+++ b/lib/fog/libvirt/models/compute/templates/server.xml.erb
@@ -60,9 +60,11 @@
       <model type='<%= nic.model %>'/>
     </interface>
 <% end -%>
+<% if guest_agent -%>
     <channel type='unix'>
       <target type='virtio' name='org.qemu.guest_agent.0'/>
     </channel>
+<% end -%>
     <serial type='pty'>
       <target port='0'/>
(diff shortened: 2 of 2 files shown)


strzibny:
I don't mind this being default if there no objections but in your patch this is not configurable.

olifre:
> I don't mind this being default if there no objections but in your patch this is not configurable.

I am unsure on how to make it configurable, since I am not really an expert of libfog and the API. Can you give me some pointers on how to add it, or add the configurability part to this PR? 
It's probably rather straightforward once one knows which parts need to be touched.

strzibny:
I believe you have to add attribute here:
https://github.com/fog/fog-libvirt/blob/master/lib/fog/libvirt/models/compute/server.rb#L14

A default value here:
https://github.com/fog/fog-libvirt/blob/master/lib/fog/libvirt/models/compute/server.rb#L465

And then use plain old ERB for inserting the value in the template:
```
<%= guest_agent %>
```

olifre:
@strzibny Thanks! I now made the guest-agent toggleable, on by default. 
Configuring the name or the complete XML snippet is not really useful IMHO, since this is the name expected by libvirt / virsh and used in default guest agent configurations. So I just added a bool which allows to turn the agent channel on or off.

olifre:
Any news on this?

olifre:
@strzibny I've now rebased on current master and manually fixed the conflicts - would be nice to get this merged.

strzibny:
Yes, I think this is fine. Thanks for your contribution!

PR #60, opened 2018-10-18
Allow to specify UID and GID of volume.

voxik:
This would allow to remove some duplicated code in vagrant-libvirt/vagrant-libvirt/pull/868


plribeiro3000:
LGTM!

PR #59, opened 2018-07-31
fog-libvirt.gemspec: changed dependency for fog-core to ~>2.1

johanneskastl:
Fixes issue #58.

Needs to be tested, if there are any regressions.

PR #61, opened 2018-11-01
Follow fog-core best practices.

voxik:
Following fog/fog-core#229, this gem needs to be namespaced.

Please note that this fixes the "libvirt has no compute service (Fog::Service::NotFound)" error as can be seen in the test result of PR #59. I have also tested this locally against fog-core 1.43 and this PR should be backward compatible.


strzibny:
This patch looks good to me, I also tried with never versions of Ruby. Some people might still work with the old constants so this should be part of `0.6.0` release at the very least.

@voxik Thanks a lot for the patch!

Note: the Travis failures are troubles with Gemfile-edge which I will try to fix.

PR #62, opened 2018-12-04
Handle blank text in allocation

apuntamb:
This PR handles the failure caused when allocation is passed empty in foreman libvirt .

### DIFF
--- a/lib/fog/libvirt/models/compute/volume.rb
+++ b/lib/fog/libvirt/models/compute/volume.rb
@@ -114,14 +114,12 @@ module Fog
         end
 
         def split_size_unit(text)
-          if text.kind_of? Integer
-            # if text is an integer, match will fail
-            size    = text
-            unit    = 'G'
-          else
-            matcher = text.match(/(\d+)(.+)/)
+          if (text.kind_of? String) && (matcher = text.match(/(\d+)(.+)/))
             size    = matcher[1]
             unit    = matcher[2]
+          else
+            size    = text.to_i
+            unit    = "G"
           end
           [size, unit]
         end


strzibny:
Will this be always an empty string? Should we also guard against allocation being `nil`?

apuntamb:
@strzibny Hi. Sure the related foreman code where I'd made changes (in progress) is here: [theforeman/foreman/#6300](https://github.com/theforeman/foreman/pull/6300).

lib/fog/libvirt/models/compute/volume.rb
```diff
@@ -118,6 +118,9 @@ def split_size_unit(text)
             # if text is an integer, match will fail
             size    = text
             unit    = 'G'
+          elsif text.blank?
```
strzibny: I think it's better to do `text.to_i` because it yields 0 for both nil or empty string and saves us an "elseif" branch reducing complexity.

lib/fog/libvirt/models/compute/volume.rb
```diff
@@ -114,14 +114,12 @@ def defaults
         end
 
         def split_size_unit(text)
-          if text.kind_of? Integer
-            # if text is an integer, match will fail
-            size    = text
-            unit    = 'G'
-          else
-            matcher = text.match(/(\d+)(.+)/)
+          if (text.kind_of? String) && (matcher = text.match(/(\d+)(.+)/))
```
apuntamb: @strzibny How does this look now?

xprazak2: Takes care of the issue for me.

strzibny:
Looks good to me, merging.

xprazak2:
thank you @strzibny! Could we get a new release with this fix please?

xprazak2:
@strzibny, any chance you will have time to cut a new release in a near future?

strzibny:
Yes, I was waiting for https://github.com/fog/fog-libvirt/pull/66 and maybe even https://github.com/fog/fog-libvirt/pull/56

PR #65, opened 2019-03-02
libvirt template: Add hugepages support

mmoll:
If the attribute 'hugepages' is set to true, define the guest as
having its memory backed by a hugepages pool.
    
closes GH-6


mmoll:
rebased and updated version of #45

PR #66, opened 2019-09-15
fog-libvirt.gemspec: remove version constraint on fog-xml

johanneskastl:
Based on the discussion in #59, remove the version constraint on fog-xml.

Please ignore the branch name, this does not remove the dependency, only the version constraint.

### DIFF
--- a/fog-libvirt.gemspec
+++ b/fog-libvirt.gemspec
@@ -27,7 +27,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency("fog-core", ">= 1.27.4")
   s.add_dependency("fog-json")
-  s.add_dependency("fog-xml", "~> 0.1.1")
+  s.add_dependency("fog-xml", "~> 0.1")
   s.add_dependency('ruby-libvirt','>= 0.7.0')
   s.add_dependency("json")
 


strzibny:
I would still keep the lowest requirement there since I do not believe it works well with all previous versions of fog-xml.

johanneskastl:
Version 1.1.3 was released 2017, should we use that? Or allow all 1.1.x version by using `~> 1.1`

strzibny:
I am for limiting it to higher than current `">= 0.1.1"` and lower than `"< 0.2"`. This way we avoid possible break for `0.2` but allow wide range of fog-xml versions (there might be people with various versions in the wild).

johanneskastl:
I have now added a constraint that should be satisfied by all 0.1.x versions, but not by 2.x.

fog-libvirt.gemspec
```diff
@@ -27,7 +27,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency("fog-core", ">= 1.27.4")
   s.add_dependency("fog-json")
-  s.add_dependency("fog-xml", "~> 0.1.1")
+  s.add_dependency("fog-xml", "~> 0.1")
```
strzibny: This allows all versions up to `1.x` if I am not mistaken including `0.1` and `0.2`. `~>` operator allows to iterate on the last specified number (expect when this is a single number).

johanneskastl: Hmm, I think you are right. So I guess this PR is kinda pointless, then.

PR #68, opened 2020-01-23
Add support for virtio-rng.

olifre:
This allows to push entropy from the KVM host into the VM. 
It is supported since many years and happens in-kernel without any need for configuration. Since gathering good entropy inside VMs is hard without this support, this strongly improves security and performance of encrypted connections.

### DIFF
--- a/lib/fog/libvirt/models/compute/templates/server.xml.erb
+++ b/lib/fog/libvirt/models/compute/templates/server.xml.erb
@@ -94,11 +94,16 @@
       <target type='virtio' name='org.qemu.guest_agent.0'/>
     </channel>
 <% end -%>
-<% if virtio_rng -%>
     <rng model='virtio'>
-      <backend model='random'/>
-    </rng>
+<%
+  rng_backend_model = virtio_rng[:backend_model] ? virtio_rng[:backend_model] : 'random'
+-%>
+<% if virtio_rng[:backend_path] -%>
+      <backend model='<%= rng_backend_model %>'><%= virtio_rng[:backend_path] %></backend>
+<% else -%>
+      <backend model='<%= rng_backend_model %>'/>
 <% end -%>
+    </rng>
     <serial type='pty'>
       <target port='0'/>
     </serial>
--- a/lib/fog/libvirt/models/compute/templates/server.xml.erb
+++ b/lib/fog/libvirt/models/compute/templates/server.xml.erb
@@ -94,6 +94,11 @@
       <target type='virtio' name='org.qemu.guest_agent.0'/>
     </channel>
 <% end -%>
+<% if virtio_rng -%>
+    <rng model='virtio'>
+      <backend model='random'/>
+    </rng>
+<% end -%>
     <serial type='pty'>
       <target port='0'/>
     </serial>
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -483,7 +483,7 @@ module Fog
             :cpu                    => {},
             :hugepages              => false,
             :guest_agent            => true,
-            :virtio_rng             => true,
+            :virtio_rng             => {},
           }
         end
 
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -30,6 +30,7 @@ module Fog
         attribute :cpu
         attribute :hugepages
         attribute :guest_agent
+        attribute :virtio_rng
 
         attribute :state
 
@@ -482,6 +483,7 @@ module Fog
             :cpu                    => {},
             :hugepages              => false,
             :guest_agent            => true,
+            :virtio_rng             => true,
           }
         end
 


olifre:
@strzibny Do you have time to review this one? 
It's significantly easier than the other PR ;-).

lib/fog/libvirt/models/compute/templates/server.xml.erb
```diff
@@ -93,6 +93,11 @@
     <channel type='unix'>
       <target type='virtio' name='org.qemu.guest_agent.0'/>
     </channel>
+<% end -%>
+<% if virtio_rng -%>
+    <rng model='virtio'>
+      <backend model='random'/>
```
strzibny: What about actually offering to provide the backend instead of relying on hypervisor default? As the documentation says "For QEMU, the default is /dev/random. However, the recommended source of entropy is /dev/urandom (as it doesn't have the limitations of /dev/random). "

olifre: That's certainly possible. However, I'm unsure which default we should then use in `libfog`. Reading through https://bugzilla.redhat.com/show_bug.cgi?id=1074464 , it seems support for `/dev/urandom` has only been added in recent years. 
Both `/dev/random` and `/dev/urandom` can be problematic, though:
- `/dev/random` blocks if emptied by rogue VMs (but at least does not break quality), which in turn may break `/dev/urandom` in VMs. 
- `/dev/urandom` does not block if there is no entropy, potentially leading to bad randomness. 

The ideal solution would be to use `/dev/urandom` when enough entropy is available, and block otherwise. This is actually discussed in:
https://bugzilla.redhat.com/show_bug.cgi?id=1689946
and will be part of libvirt 6.1.0 via the `builtin` backend, which uses the `getrandom` syscall which has exactly this behaviour (c.f. https://github.com/libvirt/libvirt/commit/51489b84194b67be0d56820485de5aca555ebe97 ). But we can't rely on this yet, I think. 

So it is probably wise to expose both the `backend model` and then also the `backend path`, I agree. The question is: What should we default to, the most compatible, current default (`/dev/random`) or the recommended, but less compatible default (`/dev/urandom`)? 

Sorry, I really thought this PR was easier, but it turns out it's not...

strzibny: We should be ideally doing both as you said. I would introduce a simple array or hash to pass both values at once (no two settings). Default could be your current version, I don't see an issue with it, but at the same time it would mean that we use the default only when the user provided settings is wrong/unsupported, so maybe there don't have to be a default at all and don't have to limit what's provided either. Thoughts?

olifre: I thought about it a bit and discussed with colleagues today. I think this is the best way to go, especially thinking about future changes to the defaults. I skimmed through `qemu` a bit and it seems they have moved to `/dev/urandom` as default [recently](https://github.com/qemu/qemu/commit/a2230bd778d8445231d2120b87be3ed262bb4f8a) and to `rng-builtin` (which uses `getrandom`) even more [recently](https://github.com/qemu/qemu/commit/0198c2621a1e57d4ceaca1326897f007963c12df). Libvirtd should automatically follow the changed defaults, so not having a default in libfog and not limiting anything seems to be the most reasonable. 
I'll give implementation and testing a go later :smile: .

olifre: @strzibny What I pushed just now is completely untested (so there could be anything from syntax to logic errors) - I only pushed it to ask if this matches what you have been thinking of before I go ahead and test it in the next days.

lib/fog/libvirt/models/compute/templates/server.xml.erb
```diff
@@ -94,6 +94,16 @@
       <target type='virtio' name='org.qemu.guest_agent.0'/>
     </channel>
 <% end -%>
+    <rng model='virtio'>
+<%
+  rng_backend_model = virtio_rng[:backend_model] ? virtio_rng[:backend_model] : 'random'
+-%>
+<% if virtio_rng[:backend_path] -%>
```
strzibny: I think it would be a little better to ask on `virtio_rng[:backend_model]` and if provided we include the `<rng>` tag with either backend/path (if both are provided) or just the backend with default path. 

I am saying that to not introduce `rng` automatically for all with the next release.

wiene: > I am saying that to not introduce `rng` automatically for all with the next release.

I naively assume that enabling `virtio_rng` is a useful default setting for people not spending any thoughts on random numbers. Therefore I would like to gain a better understanding of the reasons for your concerns.

olifre: @strzibny I tend to agree with @wiene here: Having to virtio-rng inside a VM is unexpected to those aware of the problem to get entropy inside a virtual environment, and harmful to those who are unaware, since they are subject to security and / or performance issues. 
So I would vote for {{rng}} to be added and "on by default", if that's not a too significant change for the next release. Thoughts?

strzibny: I am actually not against it. My comment was purely based on selfish maintainership issue of me spending time now to find out if it's okay as a default :). As a non-default I would already merge it. If anything it provoked a response from @olifre which gives me confidence in this being a default.

olifre: > My comment was purely based on selfish maintainership issue of me spending time now to find out if it's okay as a default :)

That's fine ;-). Better ask for more reasoning early than be sorry later. 

> If anything it provoked a response from @olifre which gives me confidence in this being a default.

So that means I've convinced you that the current proposal is fine? Then I would go ahead and give it some testing and report back.

strzibny: Yes.

olifre: Thanks! 
I gave it sufficient testing now - trying different ``backend_path``s for ``backend_model`` ``random``, also trying ``backend_model`` ``rng-builtin`` (which expectedly caused an error message by our ``libvirtd`` since it does not support that yet). It seems to work as planned for us :-).

strzibny: @olifre thanks a lot for the testing, I will merge now:)

olifre: @strzibny Thanks, and thanks for the helpful discussion, too :+1:.

PR #69, opened 2020-03-17
Being able to create cloud-init iso using mkisofs

jaxi:
MacOS does not support `genisoimage`. Given the behaviour between `mkisofs` and  `genisoimage` is identical (on CentOS 8 it's mkisofs is even symlinked to genisoimage), it will really be useful to use `mkisofs` as a fallback of `genisoimage`.

Creating domain via cloud-init using mkisofs looks like

```ruby
server = fog.servers.create(
  name: domain_name,
  cpus: 1,
  memory_size: 2048 * 1024,
  domain_type: 'kvm',
  os_type: 'hvm',
  network_nat_network: 'vm-network',
  volumes: [ new_volume ],
  user_data: user_data,
  geniso_util: 'mkisofs',
)
```

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -1,6 +1,7 @@
 require 'fog/compute/models/server'
 require 'fog/libvirt/models/compute/util/util'
 require 'fileutils'
+require 'tempfile'
 
 module Fog
   module Libvirt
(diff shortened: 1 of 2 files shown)


lib/fog/libvirt/models/compute/server.rb
```diff
@@ -1,6 +1,7 @@
 require 'fog/compute/models/server'
 require 'fog/libvirt/models/compute/util/util'
 require 'fileutils'
+require 'tempfile'
```
jaxi: Without this fog-libvirt throws `uninitialized constant Fog::Libvirt::Compute::Server::Tempfile`. Might have something to do with Ruby version... (I am on `ruby 2.6.3p62`)

PR #71, opened 2020-03-26
Update video card in server.xml.erb

qpe-y9d37y:
Changing video card from ``<model type='cirrus' vram='9216' heads='1'/>`` to ``<model type='qxl' ram='65536' vram='65536' vgamem='16384' heads='1'/>`` as 'cirrus-vga' is deprecated.
This change removes the warning message "qemu-kvm: -device cirrus-vga,id=video0,bus=pci.0,addr=0x2: warning: 'cirrus-vga' is deprecated, please use a different VGA card instead" found in the libvirt logs when starting a VM.

### DIFF
--- a/lib/fog/libvirt/models/compute/templates/server.xml.erb
+++ b/lib/fog/libvirt/models/compute/templates/server.xml.erb
@@ -137,7 +137,7 @@ end
 -%>
     <graphics type='<%= display_type %>' port='<%= display_port %>' autoport='<%= autoport %>' <%= display_listen %> <%= display_password %> />
     <video>
-      <model type='cirrus' vram='9216' heads='1'/>
+      <model type='qxl' ram='65536' vram='65536' vgamem='16384' heads='1'/>
     </video>
   </devices>
 </domain>


plribeiro3000:
@qpe-y9d37y Wont this break old versions?
Maybe we could keep it backwards compatible somehow?

lzap:
I agree, this should be ideally a new flag with VGA being default it could be flipped when 1.0.0 is released (with other breaking changes).

lzap:
We have another report that libvirt in CentOS 8.3 does not even work properly (VNC) with the old cirrus driver: https://community.theforeman.org/t/unable-to-open-the-vnc-console/23660

@plribeiro3000 what is your recommendation? I think something needs to be done here.

plribeiro3000:
Can we work towards making it backwards compatible?

I'm worried that it might be working for some and this change break it.

A simple conditional with a default for the actual value would be enough.

lzap:
I agree, I can rewrite the patch and introduce a new argument for this, @qpe-y9d37y is from our (Foreman) community and I am not sure if there is still an interest in this. We have a bugreport from other our user so this is something we would like to fix.

But before that I must ask - what it takes in your community to get push permissions? I'd be interested, there seem to be a lot of work in here. Some PRs can be merged right away. Do you have some nomination process?

lzap:
@geemus I assume you guys are busy and I would like to help:

* https://github.com/fog/fog-libvirt/pull/86 this is RTM
* https://github.com/fog/fog-libvirt/pull/87 RTM as well
* https://github.com/fog/fog-libvirt/pull/88 fixed typo, ready for review
* https://github.com/fog/fog-libvirt/pull/77 this was dropped for no feedback
* https://github.com/fog/fog-libvirt/pull/69 this is a valid PR, I could help with testing

And this PR too is something I can work with the original author finding a backward compatible solution, but as you can see, I am not very motivated to start another work since all those PRs are just hanging. What can I do to get things moving again? :-)

geemus:
@lzap thanks for the ping, this hadn't yet landed on my radar (I had largely delegated as I've never personally been a libvirt user). I'd be happy to add you to committers as a next step, sorry for the gap in service here. Just let me know if you need anything else to get this all moving again.

lzap:
Thanks for the trust, I will go ahead and merge easy PRs and poke unfinished ones.

I'd also like to get a release out of doors too, who should I ping once the tag is ready? Is there some pre-release process like checking which version of fog-gore particular version of implementation is compatible with? I guess you guys do not do semantic versioning since the library is still at 0.y.z. Any documentation in that regard would be helpful.

Let's get things moving! 👍🏻

lzap:
@qpe-y9d37y can you let me know if you are able to rewrite the patch in a way that new argument is added and by default the old video card configuration is kept but users can opt-in for the new one? We can deprecate the functionality, cut a new release and only after that we can talk about making a breaking change.

geemus:
@lzap Of course, thanks for your offer to help us get this moving again.

I try my best to follow semantic (except the major, which honestly should probably just be 1 or higher now), so continuing that would be great. 

I have release permissions, and given that the other folks have been inactive, pinging me is probably going to be best/smoothest. I don't think we have a clear pre-release process at present, it's been more ad-hoc, but maybe we should do more to make this explicit. I think for now we can just talk through any questions/concerns and proceed from there.

Similarly if you have questions or need support on particular issues, feel free to ping me and I'll do my best to answer questions and support.

lestercheung:
Hello from 2025! Latest Foreman v3.13 using `rubygem-fog-libvirt-0.12.2-1.el9.noarch` is still creating VMs with the Cirrus Logic VGA - is there any plans to switch over with or without backward compatibility? Any advice for existing users?

ekohl:
These days `video` is an attribute that can be passed in. It still defaults to `cirrus`, but can easily be changed by the caller.

Having said that, I think you're right that it's time to update the default here:

https://github.com/fog/fog-libvirt/blob/f46eac9ad6bfb3a519b5c21c03e156868436943d/lib/fog/libvirt/models/compute/server.rb#L561

geemus:
I'm certainly open to changing the default. That said, I'm not using libvirt myself, so I'm reticent to work from educated guesses. Could one of you work on a PR with the specific change and I then I can see about reviewing and releasing? Thanks!

ekohl:
https://www.kraxel.org/blog/2014/10/qemu-using-cirrus-considered-harmful/ is just over 10 years old now. Looks like `vga` is likely the best choice given that at least RHEL has stopped shipping spice and Windows requires drivers. https://libvirt.org/formatdomain.html#video-devices has more. It says you can leave out the vram and let it use a default value, which is probably better than what we come up with.

ekohl:
https://github.com/fog/fog-libvirt/pull/170

PR #73, opened 2020-04-06
Fixes #7 Add autostart feature

### LINKED ISSUE #7: autostart support for servers
Hi,

There is an `autostart` attribute in the server model for libvirt. However, I don't see where this attribute could be enabled when creating a VM. I don't know libvirt's API, but it seems there's a [separate method](https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainSetAutostart) to call when enabling `autostart`.

Is it currently possible to enable `autostart` when creating a VM with fog? If so, how? If not, would it be possible to add it?

Thank you.

Original issue opened by @eolamey at fog/fog#3343

cc/ @eolamey @geemus

apatelKmd:
This is my first iteration to receive useful feedback. I wanted to test more thoroughly but saw the migration snippets from shindo to minitest.

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -78,6 +78,10 @@ module Fog
           action_status
         end
 
+        def update_autostart(value)
+          service.update_autostart(uuid, value)
+        end
+
         def mac
           nics.first.mac if nics && nics.first
         end
@@ -132,9 +136,10 @@ module Fog
         end
 
         #alias methods
-        alias_method :halt,    :poweroff
-        alias_method :stop,    :shutdown
-        alias_method :active?, :active
+        alias_method :halt,       :poweroff
+        alias_method :stop,       :shutdown
+        alias_method :active?,    :active
+        alias_method :autostart?, :autostart
 
         def volumes
           # lazy loading of volumes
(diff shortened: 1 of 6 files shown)


lib/fog/libvirt/models/compute/server.rb
```diff
@@ -78,6 +78,18 @@ def start
...
+          service.vm_action(uuid, :get_autostart)
+        end
+
+        def autostart
+          return true if autostart?
+        end
+
+        def update_autostart(value)
+          service.vm_action(uuid, :update_autostart, value)
+        end
```
apatelKmd, line 91: Request to update the value

minitests/server/server_test.rb
```diff
@@ -6,6 +6,13 @@ def setup
     @server = @compute.servers.new(:name => "test")
   end
 
+
+  def test_autostart_success
+    assert_equal true, @server.autostart?
+    @server.update_autostart(false)
+    assert_equal false, @server.autostart?
+  end
```
apatelKmd, line 14: I want to do a real test but stuck on the Mock setup.

geemus:
@apatelKmd looks pretty good overall. Do you have any particular questions or things we can help with?

apatelKmd:
I have verified changes locally through using the `bundle exec rake console` task and modifying existing vm `autostart` attribute.

@geemus see the following:
 * I wanted to implement more natural setter method, `def autostart = value`. Not knowing Ruby and the framework I settled with what I know `def update_autostart`.

 * I am unable to fully test out `request` test. I wanted to set/reset and test out the value of `autostart` attribute but `autostart?` always returns the default `false`.

Other than the above mentioned two issues proposed changes are functional.

PR #75, opened 2020-04-23
Return empty collection if no matching volume found

electrofelix:
Change the saving of the return from get_volume to ensure the
resulting list is empty if no volume was found based on the filter
keys specified.

Update the mock to replicate similar behaviour that can be
expected to be relied upon that highlights a small number of
assumptions relying on existing mock behaviour that would not
align with real world experiences.

### DIFF
--- a/lib/fog/libvirt/models/compute/volume.rb
+++ b/lib/fog/libvirt/models/compute/volume.rb
@@ -60,6 +60,7 @@ module Fog
           new_volume.save
 
           new_volume.reload
+          new_volume
         end
 
         def clone_volume(new_name)
@@ -71,6 +72,7 @@ module Fog
           new_volume.path = service.clone_volume(pool_name, new_volume.to_xml, self.name).path
 
           new_volume.reload
+          new_volume
         end
 
         def upload_image(file_path)
--- a/lib/fog/libvirt/models/compute/volume.rb
+++ b/lib/fog/libvirt/models/compute/volume.rb
@@ -60,7 +60,6 @@ module Fog
           new_volume.save
 
           new_volume.reload
-          new_volume
         end
 
         def clone_volume(new_name)
@@ -72,7 +71,6 @@ module Fog
           new_volume.path = service.clone_volume(pool_name, new_volume.to_xml, self.name).path
 
           new_volume.reload
-          new_volume
         end
 
         def upload_image(file_path)
--- a/lib/fog/libvirt/requests/compute/list_volumes.rb
+++ b/lib/fog/libvirt/requests/compute/list_volumes.rb
@@ -78,15 +78,22 @@ module Fog
         def list_volumes(filters={ })
           vol1 = mock_volume 'vol1'
           vol2 = mock_volume 'vol2'
-          [vol1, vol2]
+          vols = [vol1, vol2]
+
+          if filters.keys.empty?
+            return vols
+          end
+
+          key = filters.keys.first
+          vols.select { |v| v[key] == filters[key] }
         end
 
         def mock_volume name
           {
               :pool_name   => 'vol.pool.name',
               :key         => 'vol.key',
-              :id          => 'vol.key',
-              :path        => 'vol.path',
+              :id          => "vol.#{name}",
+              :path        => "path/to/disk", # used by in mock_files/domain.xml
               :name        => name,
               :format_type => 'raw',
               :allocation  => 123,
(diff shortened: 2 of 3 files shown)


lib/fog/libvirt/models/compute/volume.rb
```diff
@@ -71,6 +72,7 @@ def clone_volume(new_name)
           new_volume.path = service.clone_volume(pool_name, new_volume.to_xml, self.name).path
 
           new_volume.reload
+          new_volume
```
electrofelix, line 75: I'm don't think this is correct, during modifying the tests i discovered that `new_volume.reload` wasn't returning a volume here and I suspect I'm missing understanding something around the mocks behaviour as to why reload returns nil here when I adjust what volumes are returned.

Same for the above, I just modified both to be consistent.

lib/fog/libvirt/requests/compute/list_volumes.rb
```diff
@@ -69,23 +69,31 @@ def get_volume filter = { }, raw = false
...
         end
 
         def mock_volume name
           {
               :pool_name   => 'vol.pool.name',
               :key         => 'vol.key',
-              :id          => 'vol.key',
-              :path        => 'vol.path',
+              :id          => "vol.#{name}",
+              :path        => "path/to/disk", # used by in mock_files/domain.xml
```
electrofelix, line 96: one of the tests results in searching for a volume with the path set to the same value in the domain.xml mock_files, with filtering enabled, this seemed like the easiest solution.

electrofelix:
@geemus wondering if you can help indicate how to correct the tests around the volume reload behaviour?

lib/fog/libvirt/models/compute/volume.rb
geemus: I believe reload will basically do `volumes.get(new_volume.identity)` in this context and should return the result of that call. You can see that here: https://github.com/fog/fog-core/blob/7865ef77ea990fd0d085e49c28e15957b7ce0d2b/lib/fog/core/model.rb#L44

Does that help clarify what the expectations would be? I'm not familiar with the current state of the mocks and things, but hope this might help point you in the right direction. I'm certainly happy to answer other questions as you have them also.

electrofelix: Thanks, I think it helped give me the hint that reload was calling list_volumes with a filter and that allowed me to work out that reload will trigger a look up of the volume by `key` using the volume `:id` attribute. So in the mocks below, to support filtering it is necessary to ensure `:id` and `:key` have the same value.

geemus: Great, glad that helped nudge you in the right direction.

electrofelix:
@geemus anything else needed for this to merge?

lib/fog/libvirt/requests/compute/list_volumes.rb
```diff
@@ -69,23 +69,31 @@ def get_volume filter = { }, raw = false
               return raw ? vol : volume_to_attributes(vol)
             end
           end
-          { }
+
+          nil
```
geemus, line 73: The rest seems reasonable, but I'm not sure about this particular change. It doesn't seem related to the other changes, and I fear it could break existing code/usage. Am I missing something and/or could we omit this part? Thanks!

geemus (changes requested):
Looks pretty close, I asked one question for us to discuss. Also, apologies for not replying earlier, I must have missed that notification somehow.

lib/fog/libvirt/requests/compute/list_volumes.rb
electrofelix: @geemus isn't this method private to the class?

The change here is to avoid adding empty hashes to the list_volumes response above. I guess the question would be if code would be relying on whether list_volumes always returns a list of empty hashes?

geemus: Looking again, I think this should probably be ok. Especially as it is private, I think maybe I overlooked that fact when thinking about it before.

geemus (approved):

lib/fog/libvirt/requests/compute/list_volumes.rb
electrofelix: You know what, I should have added a test, it turns out that while this avoids adding a list of hashes, the code above doesn't skip appending a nil entry

electrofelix: And on further examination, that's probably alright because volume_to_attributes will return nil entries, so at least anyone's code at this point need only handle skipping nil entries.

PR #77, opened 2020-06-24
template changes to allow for xen domain_type  vm creation

alvinstarr:
This change will create a xen domain once the domain_type is set to xen from kvm.
This works within Foreman and hopefully once i edit the server.rb file.
I am not familiar enough with the inner workings of Foreman to know where to alter the domain_type.


lzap:
There are conflicts now, I do not have a xen setup, but I wouldn't mind testing for regressions on Linux and reviewing the code.

There seems to be lack of man power for this repo, we can probably help. I am from Red Hat working in Satellite team and we have a dependency on this package. I think we can find one or two engineers who could jump aboard helping with reviews, merging, putting together releases more often.

PR #78, opened 2020-12-01
Fixes #24 - don't fail if bridge is undefined

### LINKED ISSUE #24: Support for hostdev mode network
Hi,
I have libvirt network with hostdev mode (pool of virtual NICs on PCI-E) and server is managed by Foreman.
This type of network seems to be unsupported by fog-libvirt.

&lt;network&gt;
  &lt;name&gt;pf-eth0&lt;/name&gt;
  &lt;uuid&gt;0ae0eb24-d0d9-4a02-942e-d245c77e12ba&lt;/uuid&gt;
  &lt;forward mode='hostdev' managed='yes'&gt;
    &lt;driver name='vfio'/&gt;
    &lt;pf dev='eth0'/&gt;
  &lt;/forward&gt;
&lt;/network&gt;

Backtace from foreman is attached as [trace.txt](https://github.com/fog/fog-libvirt/files/467088/trace.txt)

Is someone willing to add such support please?

ares:


lib/fog/libvirt/requests/compute/list_networks.rb
```diff
@@ -25,12 +25,14 @@ def get_network_by_filter(filter)
           end
         end
 
+        # bridge name may not be defined in some networks, we should skip that in such case
         def network_to_attributes(net)
           return if net.nil?
+          bridge_name = net.bridge_name rescue ''
```
plribeiro3000, line 31: Does `bridge_name` returns an exception?

plribeiro3000: Could we rescue only the exception raised by it instead of silently swallow any error?

ares: it does, sadly it's quite generic - `Libvirt::Error` but you're right, it's better to specify it explicitly

ares:
PR updated

PR #79, opened 2020-12-11
Skip monitor loop if it's not defined in ceph RBD disks

GermanG:
It's already defined in the pool.
We have multiple ceph clusters in different environments, so we have many monitor groups.
This way we let the configuration manager define the pool with its mons and keep it simple.
The only issue is if you want to name the pool differently per environment and
if you want to have more than a rbd pool per host.


plribeiro3000 (approved):

plribeiro3000:
LGTM

PR #82, opened 2021-02-11
Add support for s390x architecture

lzap:
There is no USB, PS2 or graphics output on IBM Z. The console must be configured for SCLP target in order to be able to connect to the guest.

### DIFF
--- a/lib/fog/libvirt/models/compute/templates/server.xml.erb
+++ b/lib/fog/libvirt/models/compute/templates/server.xml.erb
@@ -110,6 +110,13 @@
       <backend model='<%= rng_backend_model %>'/>
 <% end -%>
     </rng>
+<% if arch == "s390x" -%>
+    <controller type="scsi" index="0" model="virtio-scsi"/>
+    <console type="pty">
+      <target type="sclp"/>
+    </console>
+    <memballoon model="virtio"/>
+<% else -%>
     <serial type='pty'>
       <target port='0'/>
     </serial>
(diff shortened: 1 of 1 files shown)


lib/fog/libvirt/models/compute/templates/server.xml.erb
```diff
@@ -118,6 +119,14 @@
     </console>
     <input type='tablet' bus='usb'/>
     <input type='mouse' bus='ps2'/>
+<% end -%>
+<% if arch == "s390x" -%>
```
plribeiro3000, line 123: Could you avoid the extra condition check here and do an `if else`?

lzap: Asbolutely, good point. I am actually reversing the order so it's more readable: if s390x else.

lzap:
I rebased and amended the requested change, apologies for the delay.

lzap:
Can I ask you for a favor to cut a release with this patch? I have a customer who would love to see this :-) Cheers.

lib/fog/libvirt/models/compute/templates/server.xml.erb
plribeiro3000: 🆒

plribeiro3000 (approved):

plribeiro3000:
@lzap Actually, could you wait a bit? We just did a release and your PR is the first to get merged.
We already have a couple PR's waiting some answers to be merged.

lzap:
Absolutely, thank you so much! Cheers.

lzap:
Hey, any chance to get a release out? :-)

PR #83, opened 2021-02-11
Minimum Ruby version is 2.4

lzap:
- Drop older and unsupported Ruby versions up to 2.3.x.
- Remove committed .ruby-version file and add it to gitignore
- Update Rubocop configuration so modern Rubocop can actually work


plribeiro3000:
@lzap You are doing several structural refactorings over here. Could you split them in different PR's or at least in different commits?

I have some concerns about your changes:
-  Update rubocop just for the sake of update and add everything as something safe to be fixed later is not a good approach as those are gonna get forgotten
- We need to be sure about the version dependency as `fog` still support `2.0.0+` because of some constraints we have with old linux distros in LTS
- I personally do like `.ruby-version` and `.ruby-gemset` to be in the project so i don't have to worry about it every time i have to pull the project again

Can we split all of those and talk about them separately?

lzap:
It's fine, was just a suggestion, looks like you don't like it all so I will just close it.

PR #84, opened 2021-02-11
Fix shell escape vulnerability

lzap:
Older versions of libvirt (version 1002008 or order) or builds without dhcp leases feature compiled it are vulnerable to shell escape through for-libvirt. An attacker can execute arbitrary shell code either locally or remotely over ssh.

Luckily, most linux distributions these days have the dhcp leases feature compiled in and libvirt library in version that prevents from the affected code being executed.

This bug still deserves a fix if there's someone running an old hypervisor or calling those methods explicitly via send.

The following code prints "test" string into system log/journal by crafting a domain name to escape the shell:

```ruby
require "fog/libvirt"
compute = Fog::Compute.new(provider: :libvirt, libvirt_uri: "qemu:///system")
server = compute.servers.create(name: "test; logger test")
# Use this call to trigger the bug on systems with
# recent version of libvirt:
server.send(:addresses_ip_command)
```

The proposed fix is very simple so it could be backported if you choose to do so. I suggest to drop the code in question completely from the future versions of fog-libvirt.

Red Hat Security Response Team was made aware of this bug, no products are known to be affected, no CVE was filed.

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -347,6 +347,9 @@ module Fog
           ip_address.chomp
         end
 
+        # Locale-friendly removal of non-alpha nums
+        DOMAIN_CLEANUP_REGEXP = Regexp.compile('[\W_-]')
+
         # This retrieves the ip address of the mac address using ip_command
         # It returns an array of public and private ip addresses
         # Currently only one ip address is returned, but in the future this could be multiple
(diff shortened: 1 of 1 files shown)


ares (approved):
Makes sense

lib/fog/libvirt/models/compute/server.rb
```diff
@@ -360,7 +360,7 @@ def addresses_ip_command(service_arg=service, options={})
           ip_command_global=service_arg.ip_command.nil? ? 'grep $mac /var/log/arpwatch.log|sed -e "s/new station//"|sed -e "s/changed ethernet address//g" |sed -e "s/reused old ethernet //" |tail -1 |cut -d ":" -f 4-| cut -d " " -f 3' : service_arg.ip_command
           ip_command_local=options[:ip_command].nil? ? ip_command_global : options[:ip_command]
 
-          ip_command="mac=#{mac}; server_name=#{name}; "+ip_command_local
+          ip_command="mac=#{mac}; server_name=#{name.gsub(/[\W_-]/, '_')}; "+ip_command_local
```
plribeiro3000, line 363: Alright, Just so we don't get lost (Regexp are always complex to read and follow) What do you think of saving this value in a different variable with a more suggested name?

lzap: Sure I can do that, amended the change.

plribeiro3000 (approved):

PR #87, opened 2021-05-07
Make volume owner and group optional

ekohl:
Always passing this makes it required. This breaks using libvirt via qemu:///session. By not passing it at all, it leaves it up to libvirt to deal with.


lzap (approved):
Yeah thanks for the patch, this was bugging me too.

ekohl:
Merging my own PR given the prior approval by @lzap (but then he didn't have merge access yet).

PR #92, opened 2021-06-04
Remove .ruby-{gemset,version}

ekohl:
These are used to configure RVM, but as a library this should have no opinion about that.


lzap (approved):
Yeah let's get rid of this.

Issue #76, opened 2020-06-05
Q35 chipset support?


stove-panini:
Hi, I'm using this gem by way of Foreman's libvirt provider. I'm having trouble figuring out a non-hacky way to create a machine based on the newer Q35 chipset. Currently, I'm simply editing [server.xml.erb](https://github.com/fog/fog-libvirt/blob/master/lib/fog/libvirt/models/compute/templates/server.xml.erb) and adding `machine='q35'` into the `<os><type>` key and switching the IDE cdrom for a SATA one.

I believe that changes to server.xml.erb are necessary insofar as replacing the Q35-incompatable IDE bus, but I'm having trouble finding a way to specify a machine's chipset from libvirt's Ruby API.

Any suggestions so that I could possibly get a real PR going?

plribeiro3000:
@stove-panini What exactly are you not finding? How to get the current chipset through `fog-libvirt` or how to pass the value to the `server.xml.erb` file?

stove-panini:
@plribeiro3000 I'd like to pass the value to `server.xml.erb`

plribeiro3000:
Hmmm. I don't blame you. it is kind of a mess here.

So all models include `Fog::Libvirt::Util` which should be defined at `fog/libvirt/util.rb` but is instead defined at `fog/libvirt/models/compute/util/util.rb`.

This mixin has a method called `to_xml` defined [here](https://github.com/fog/fog-libvirt/blob/fa5e98aa0270786025fa64d3a21efd66e1907fd8/lib/fog/libvirt/models/compute/util/util.rb#L19-L25):

```ruby
def to_xml template_name = nil
  # figure out our ERB template filename
  erb = template_name || self.class.to_s.split("::").last.downcase
  path = File.join(File.dirname(__FILE__), "..", "templates", "#{erb}.xml.erb")
  template = File.read(path)
  ERB.new(template, nil, '-').result(binding)
end
```

As you can see it pass the current object to the template so the `ERB` template will have [all attributes](https://github.com/fog/fog-libvirt/blob/fa5e98aa0270786025fa64d3a21efd66e1907fd8/lib/fog/libvirt/models/compute/server.rb#L14-L30) accessible as parameters.

ekohl:
I opened https://github.com/fog/fog-libvirt/pull/94 which should make it easier to fix this.

jhoblitt:
Could this issue be reopened? This is needed for running EL9 VMs.

[...]

PR #94, opened 2021-06-04
Replace server XML with Nokogiri::XML::Builder

ekohl:
This replaces the ERB template with a real XML builder. This makes it much easier to maintain.

https://github.com/fog/fog-libvirt/issues/76 inspired this change. It would make things more flexible.

While I haven't tested this very extensively, I do believe it's ready for review. If we agree this is a good path, we can fully remove the ERB templates.


lzap:
I am not a huge fan of XML builders, but I am okay with this. However I currently do not feel comfortable merging this without fixture tests. Could you add at least one fixture? Basically a rendered XML committed to git, so everytime we make a change to the builder we can compare the generated XML files. Ideally few of them for various major code paths.

There is currently `lib/fog/libvirt/requests/compute/mock_files/domain.xml` but I don't see a test for that, perhaps some leftover?

jhoblitt:
The gha logs are gone due to age.  Could we reopen this PR and rebase it to trigger gha?

ekohl:
@jhoblitt looks like the whole RuboCop config generally just doesn't pass and needs a big update.

As for this PR itself: it would be good to have some tests on this, but I don't have time for it myself anytime soon.

jhoblitt:
@ekohl To being able to create el9 VMs via foreman is causing me major problems.  What is the fastest path forward? Should I try to get this PR merged and then add machine type support to it or should I add it to the erb template?

jhoblitt:
@ekohl @lzap Not being able to set `pc-q35-rhel9.0.0` is killing me. Could we come up with a plan for a way forward?

ekohl:
I've been on vacation for most of last month. I just rebased this with some RuboCop items addressed and tests added (that caught some bugs). I'm not sure I agree with the too high complexity issues: they were already there in the template and I think this is more readable. I could split it into multiple helpers, but I wonder if that really helps.

@jhoblitt would you have the capacity to test this PR?

jhoblitt:
@ekohl Yes, I'm happy to help test.  I suppose a real-world example would be to install this branch in the foreman instance for my dev env, which does have a number of libvirt compute resources configured. However, this is still an EL7 instanced (I know, I know, I'm trying to migrate) and is currently using `0.9`.
```
tfm-rubygem-fog-libvirt-0.9.0-1.el7.noarch
```

jhoblitt:
I think I may have seen an issue with setting boot order.

When creating a new VM, I expected to get:

```
  <os>
    <type arch='x86_64' machine='pc-i440fx-rhel7.6.0'>hvm</type>
    <boot dev='network'/>
    <boot dev='hd'/>
  </os>
```

but got 

```
  <os>
    <type arch='x86_64' machine='pc-i440fx-rhel7.6.0'>hvm</type>
    <boot dev='hd'/>
  </os>
```

I discovered there were some issues with broken plugins installed in the instance and needed to do a cleanup/reinstall.  I will retest soon.

ekohl:
Ah, I think I forgot the boot order. Added.

jhoblitt:
Tomorrow is a major US holiday and I'll be away from the office but I'll be back on Wednesday and can do any needed testing then.

ekohl:
I've noticed that RuboCop is configured to prefer double quotes but a lot of the code is in single quotes. I'm wondering if we should disable that cop instead.

jhoblitt (approved):
6db6f8089418599044be39373f3baf17402b5cca has been tested as working with a live foreman instance and it was possible to create a new EL8 VM using a libvirt compute resource.

ekohl:
Unless other maintainers disagree I'm going to merge my own PR given @jhoblitt confirmed it works.

Issue #12, opened 2016-01-28
Cannot list inactive pools


electrofelix:
https://github.com/fog/fog-libvirt/blob/master/lib/fog/libvirt/requests/compute/list_pools.rb#L30 makes a call to get the num_of_volumes, however for inactive pools this will throw an exception in pool_to_attrbiutes that gets swallowed by either find_pool_by_name or find_pool_by_uuid.

I installed the gem, created a new pool in libvirt using virt-manager or virsh. Next I deactived the pool and ran the following code (similar to what vagrant-libvirt would do):
```
require 'fog/libvirt'
conn_attr = {
    :provider => "libvirt",
    :libvirt_uri => "qemu:///system",
    :libvirt_ip_command => " awk \"/$mac/ {print \\$1}\" /proc/net/arp"
}
conn = Fog::Compute.new(conn_attr)
pool = conn.client.lookup_storage_pool_by_name("test")
pool.num_of_volumes
```
The result was be similar to:
```
Libvirt::RetrieveError: Call to virStoragePoolNumOfVolumes failed: Requested operation is not valid: storage pool 'test' is not active
	from (irb):63:in `num_of_volumes'
	from (irb):63
	from /home/baileybd/.rvm/rubies/ruby-2.0.0-p643/bin/irb:12:in `<main>'
```

Looking at the states supported by list_pools, it suggests it can handle inactive pools.

This caused pradels/vagrant-libvirt#536

lzap:
Unable to reproduce with vagrant-2.2.4-2.fc30 both active and inactive are listed correctly now.

electrofelix:
I'd forgotten this issue.

The fix appears to be a relatively simple one in changing to check if the pool is active before retrieving the number of volumes:
```
--- lib/fog/libvirt/requests/compute/list_pools.rb
+++ lib/fog/libvirt/requests/compute/list_pools.rb
@@ -27,7 +27,7 @@ module Fog
             :name           => pool.name,
             :allocation     => pool.info.allocation,
             :capacity       => pool.info.capacity,
-            :num_of_volumes => pool.num_of_volumes,
+            :num_of_volumes => pool.active? ? pool.num_of_volumes : nil,
             :state          => states[pool.info.state]
           }
         end
```

Question is, what's the impact?

[...]

PR #95, opened 2021-06-10
Use a filter to allow listing of inactive pools

electrofelix:
Allow for an `:include_inactive` filter attribute to determine whether
to include or exclude inactive pools and handle correctly setting the
number of volumes in the case it is inactive.

Current testing approach requires modifying the internal function in
order to allow the Mock to call it to ensure it is exercised correctly.

Fixes: #12

### DIFF
--- a/lib/fog/libvirt/requests/compute/list_pools.rb
+++ b/lib/fog/libvirt/requests/compute/list_pools.rb
@@ -50,16 +50,9 @@ module Fog
 
       class Mock
         def list_pools(filter = { })
-          @pool_data.map do |pool|
-            Compute::Real.send(:pool_to_attributes, pool, filter[:include_inactive])
-          end.compact
-        end
-
-        def initialize(options = { })
-          @pool_data = [
-            FakePool.new(mock_pool('pool1')),
-            FakePool.new(mock_pool('pool1'))
-          ]
+          pool1 = mock_pool 'pool1'
+          pool2 = mock_pool 'pool1'
+          [pool1, pool2]
         end
 
         def mock_pool name
@@ -69,53 +62,11 @@ module Fog
               :autostart      => true,
               :active         => true,
               :name           => name,
-              :info           => {
-                :allocation     => 123456789,
-                :capacity       => 123456789,
-                :state          => 2 # running
-              },
-              :num_of_volumes => 3
+              :allocation     => 123456789,
+              :capacity       => 123456789,
+              :num_of_volumes => 3,
+              :state          => 2 # running
           }
-        end
-
-        def add_pool(pool_attributes)
-          @pool_data.append(FakePool.new(pool_attributes))
-        end
-
-        class FakePool < Fog::Model
-          # Fake pool object to allow exercising the internal parsing of pools
-          # returned by the client queries
-          identity :uuid
-
-          attribute :persistent
-          attribute :autostart
-          attribute :active
-          attribute :name
-          attribute :num_of_volumes
-          attr_reader :info
-
-          class FakeInfo < Fog::Model
-            attribute :allocation
-            attribute :capacity
-            attribute :state
-          end
-
-          def initialize(attributes = { })
-            @info = FakeInfo.new(attributes.delete(:info))
-            super(attributes)
-          end
-
-          def active?
-            active
-          end
-
-          def autostart?
-            autostart
-          end
-
-          def persistent?
-            persistent
-          end
         end
       end
     end
--- a/lib/fog/libvirt/requests/compute/list_pools.rb
+++ b/lib/fog/libvirt/requests/compute/list_pools.rb
@@ -5,19 +5,22 @@ module Fog
         def list_pools(filter = { })
           data=[]
           if filter.key?(:name)
-            data << find_pool_by_name(filter[:name])
+            data << find_pool_by_name(filter[:name], filter[:include_inactive])
           elsif filter.key?(:uuid)
-            data << find_pool_by_uuid(filter[:uuid])
+            data << find_pool_by_uuid(filter[:uuid], filter[:include_inactive])
           else
             (client.list_storage_pools + client.list_defined_storage_pools).each do |name|
-              data << find_pool_by_name(name)
+              data << find_pool_by_name(name, filter[:include_inactive])
             end
           end
           data.compact
         end
 
         private
-        def pool_to_attributes(pool)
+
+        private_class_method def self.pool_to_attributes(pool, include_inactive = nil)
+          return nil unless pool.active? || include_inactive
+
           states=[:inactive, :building, :running, :degrated, :inaccessible]
           {
             :uuid           => pool.uuid,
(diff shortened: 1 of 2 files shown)


geemus:
I'm not really very familiar with the libvirt portions of this, so a little unclear on specifically where the mock behavior is in question. That being said, fog-local might be a good example of the approach we've used elsewhere (since it is local file writes in real mode, it should hopefully be easy to follow and understand vs needing to know more about a specific provider). You can see how the mock data is stored/used here: https://github.com/fog/fog-local/blob/master/lib/fog/local/storage.rb. Happy to discuss more too, but hopefully that should give you some foundation to learn from.

electrofelix:
@geemus the problem is in testing the current behaviour of the following code https://github.com/fog/fog-libvirt/blob/64b43373722539f89458a36c02f01feb057b2a39/lib/fog/libvirt/requests/compute/list_pools.rb#L19-L46

The mock behaviour essentially by-passes testing this by returning a list of mock pools in https://github.com/fog/fog-libvirt/blob/64b43373722539f89458a36c02f01feb057b2a39/lib/fog/libvirt/requests/compute/list_pools.rb#L49-L53
So when writing tests it's not clear how to test changes to the functionality of the `pool_to_attributes` function.

The result of the `pool_to_attributes` function looks to fit into the fog model of a pool https://github.com/fog/fog-libvirt/blob/64b43373722539f89458a36c02f01feb057b2a39/lib/fog/libvirt/models/compute/pool.rb, which is fine, except what really needs to be tested is how the function handles converting a libvirt pool to a fog pool model and then how they are presented when listing all pools.

I see from the local storage that one possible solution is to duplicate the `pool_to_attributes` function in both the mock and real object:
Mock
https://github.com/fog/fog-local/blob/bf609ca9ad618dc2bf2e14013e9bb06a322e153d/lib/fog/local/storage.rb#L57-L62

Real
https://github.com/fog/fog-local/blob/bf609ca9ad618dc2bf2e14013e9bb06a322e153d/lib/fog/local/storage.rb#L89-L94

I was looking to avoid this as it requires making sure you keep them in sync, additionally the only way I could think of testing the conversion to a fog model was to construct a fake pool object that would look like what is returned by ruby-libvirt for a pool but with different internal data controlled by the test.

It feels like the mock class in https://github.com/fog/fog-libvirt/blob/master/lib/fog/libvirt/requests/compute/list_pools.rb is intended to be used by everything else that needs to call list_pools, and tests for the list pools request should be somehow using the mocked client at https://github.com/fog/fog-libvirt/blob/64b43373722539f89458a36c02f01feb057b2a39/lib/fog/libvirt/compute.rb#L55-L73

but I don't see any examples of how to accomplish this as calls to list_pools in the requests test appear to go to the mock request rather than the real request with a mocked client:
https://github.com/fog/fog-libvirt/pull/95/files#diff-be4f38ac759613aedbdee5156ed3d603cb91ab112054e6aa817da3baa2d08237R5-R9


If I was to use the fog-local as a comparison, how does the copy_object code get tested which is called at:
https://github.com/fog/fog-local/blob/8832480bd0400df6208c0aa5a1ae1cc1c640bf45/lib/fog/local/models/file.rb#L37-L42

and appears to be defined at https://github.com/fog/fog-local/blob/master/lib/fog/local/storage.rb#L82-L87

It's not clear to me that it is currently exercised by the tests?

geemus:
@electrofelix Thanks for the clarification, I think I have a much better idea of the question now.

I think there might actually be two different questions here we can tease apart. My hope is that by splitting this out a bit it will maybe make it easier to discuss (and implement), but definitely let me know if I'm still confusing things.

First, the core of your question, the pool_to_attributes stuff. I think this actually is more like some other fog tests that start with a known dataset in the format returned from the provider and use a method such as this and check that the output matches what we would expect. I think for this it probably makes sense to just do this directly with at least one sample dataset (but you could do more if there are particular edge cases or other things you felt were worth testing.

Second, the somewhat separate question of the list pool mock implementation. It definitely doesn't provide much flexibility as it stands. In some of the other fog providers we get around this by having things like list_pool look up in the stored mock data (and other calls can then update that data, so for instance creating pools would change the list). Implementing mock in this way is certainly more complex, but also brings it more inline with real usage (which makes the mocks more useful).

electrofelix:
hoping to get back to this next month

electrofelix:
Still haven't gotten to it, but hopefully soon, still on my radar though

electrofelix:
@geemus I finally got back to tweaking this, I think the main stumbling block though is the need to test the behaviour of something that is private but without having any easy way to access it via a public interface.

I've put up an adjusted patch based on my original approach, and then added a commit on top that limits the changes to having the predefined pools in the main mock and calling the private method from the tests to check it's behaviour. I'm not sure it's actually cleaner since it means knowledge of how the private method works is now directly referenced from the tests.

Please let me know if the subsequent commit is better or worse

geemus:
@electrofelix Thanks for taking another look. As you suggest, there is some stumbling bits due to how all of this fits together. I think given that, your approach seems reasonable. Just let me know if you have further questions or would like to talk through it more.

electrofelix:
@geemus which approach do you prefer? With or without the second commit?

geemus:
@electrofelix - In other fog providers for things like this, I would tend to expect one to call something like "create_pool" and have that populate mocks. Am I correct in guessing that pools are not something you can directly control from libvirt? I don't see related commands, so I'm guessing this may be something that is handled in some other more direct way? Beyond that, I guess I lean a bit toward providing some way within mocks to manipulate pools if possible, because this seems more versatile. Other people working with/on things using this might also want to change pool settings within their own mocked tests and this could allow both uses (vs only doing in tests). Does that make sense?

electrofelix:
The libvirt provider only accepts the raw XML for creating pools https://github.com/fog/fog-libvirt/blob/085a7c89596e989921dccd33dd954afbb37df27a/lib/fog/libvirt/requests/compute/define_pool.rb#L10-L12

Are you just looking for the mock to have an add_pool method and its own pool data along with a reset function?
Should I leave all the fakepool handling code with the mock then and not with the test?

electrofelix:
@geemus I just realised that with the subsequent change to move the FakePool to the tests code there is no need to modify the pools for the tests. It's not clear whether if additional functionality that would not be used by the test should be added in the same PR?

For this bug really the only thing that is of interest is to validate that `pool_to_attributes` correctly handles converting inactive pools to attributes to be subsequently returned by list_pools. Ideally it would also be nice to validate the pass thru in the `list_pools` method on the `Real` class since it contains more logic than just processing a request. Possibly there is some restructuring needed here but it's unclear to me what it should look like.

It would be possible to support additional FakePool objects being added based on a list of attributes being provide, which would then be passed through the `pool_to_attributes` method when `list_pools` is called on the mock. However it's also possibly less complex to have a method add a pool that contains the resolved attributes instead, in which case it wouldn't be used to validate the bug fix.

I can still add this, I just want to be sure that it's non usage by the test to validate the functional changes to `pool_to_attributes` makes sense and isn't going to be a cause for concern for adding excess code changes.

geemus:
@electrofelix Yeah, I think that makes sense. Lets keep it simpler and more focused as you suggest in the current iteration. Thanks for working through this with me.

geemus (approved):

PR #96, opened 2021-06-16
Cleanup test directory

lzap:
Moved tests into tests directory, both shindo and minitest are there, integration tests will go there too (TBD).


ekohl (commented):
I wonder if it makes sense to convert the minitest tests to shindo instead. For now please resolve the merge conflict though.

lzap:
I can do that probably, there is just a handful of them.

PR #102, opened 2021-07-30
Add flags option to server.destroy

byonoy-dne:
A call to `server.destroy` ultimately ends in a call to `virDomainUndefineFlags`. Right now there is no way to set the flags argument with which `virDomainUndefineFlags` gets called. This patch provides such an option to the caller.

This is needed to address [vagrant-libvirt#1027](https://github.com/vagrant-libvirt/vagrant-libvirt/issues/1027). Specifically, vagrant-libvirt needs to be able to set the `VIR_DOMAIN_UNDEFINE_NVRAM` flag when calling `destroy` on a domain configured with NVRAM.

There is one gotcha to this. If libvirt-ruby is built without `HAVE_VIRDOMAINUNDEFINEFLAGS` the call does not end in `virDomainUndefineFlags` but in `virDomainUndefine`. In that case a non-zero flags argument will result in an error. I have found no documentation or reference to that build flag. All I know for sure is that the package in the Fedora 34 repo is built with it enabled, and the user who created aformentioned issue in vagrant-libvirt also has such a build. But given that `virDomainUndefineFlags` was implemented [seven years ago](https://gitlab.com/libvirt/libvirt-ruby/-/commit/59e749fca594fdaa29bd4b3d34dfac00308286f9), it is probably ok?

Beware, this is my first contact with ruby. I hope I didn't do anything completely stupid...

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -90,9 +90,13 @@ module Fog
           volumes.first.path if volumes and volumes.first
         end
 
-        def destroy(options={ :destroy_volumes => false})
+        def destroy(options={ :destroy_volumes => false, :flags => 0 })
           poweroff unless stopped?
-          service.vm_action(uuid, :undefine)
+          if options[:flags].zero?
+            service.vm_action(uuid, :undefine)
+          else
+            service.vm_action(uuid, :undefine, options[:flags])
+          end
           volumes.each { |vol| vol.destroy } if options[:destroy_volumes]
           true
         end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -18,3 +18,6 @@ SignalException:
   Description: 'Do not enforce use of fail when raising exceptions.'
   # Valid values are: semantic, only_raise and only_fail
   EnforcedStyle: only_raise
+
+Metrics/ClassLength:
+  Enabled: false
--- a/lib/fog/libvirt/requests/compute/vm_action.rb
+++ b/lib/fog/libvirt/requests/compute/vm_action.rb
@@ -2,15 +2,15 @@ module Fog
   module Libvirt
     class Compute
       class Real
-        def vm_action(uuid, action)
+        def vm_action(uuid, action, *params)
           domain = client.lookup_domain_by_uuid(uuid)
-          domain.send(action)
+          domain.send(action, *params)
           true
         end
       end
 
       class Mock
-        def vm_action(uuid, action)
+        def vm_action(uuid, action, *params)
           true
         end
       end


lib/fog/libvirt/models/compute/server.rb
```diff
@@ -90,9 +90,9 @@ def disk_path
           volumes.first.path if volumes and volumes.first
         end
 
-        def destroy(options={ :destroy_volumes => false})
+        def destroy(options={ :destroy_volumes => false, :flags => 0 })
           poweroff unless stopped?
-          service.vm_action(uuid, :undefine)
+          service.vm_action(uuid, :undefine, options[:flags])
```
lzap, line 95: Can you pass the flags only if they are non-null? That would be backward compatible with any version. Not that this would be a huge problem, 7 years is enough in IT but...

Sorry for the delay.

lzap: My remark still stands, please just pass flags only if they are present. I cannot guarantee that passing flags: 0 will work against hypervisor running on older RHEL7.

lzap:
For the record the similar patch is here: https://github.com/fog/fog-libvirt/pull/107

The difference is that this one uses hash to pass options, which seems more robust to me.

lib/fog/libvirt/models/compute/server.rb
byonoy-dne: Sorry, I completely forget about this PR.

lzap (approved):
WFM, when tests are passing let's merge.

lzap:
Damn:

```
  For more information: https://docs.rubocop.org/rubocop/versioning.html
  reviewdog: found at least one result in diff
  reviewdog: This GitHub token doesn't have write permission of Review API [1], 
  so reviewdog will report results via logging command [2] and create annotations similar to
  github-pr-check reporter as a fallback.
```

I guess it was me who broke this. https://github.com/fog/fog-libvirt/pull/106

lzap:
Oh it was just a warning, please turn off this cop:

```
  Error: [rubocop] reported by reviewdog 🐶
  Class has too many lines. [376/372]
  
  Raw Output:
  C:  8:  7: Metrics/ClassLength: Class has too many lines. [376/372]
```

I hate mertics and style alltogether, maybe to disable them all? @ekohl

byonoy-dne:
> Oh it was just a warning, please turn off this cop:

Is this addressed to me or @ekohl? Should I make this part of this PR?

ekohl:
> I hate mertics and style alltogether, maybe to disable them all? @ekohl

It kind of depends. Generally I like to think Rubocop has mostly sane defaults, but not for everything. Perhaps we should disable `Metrics/ClassLength` in that file for now?

lzap:
@byonoy-dne just make Rubocop happy, you may disable that pesky rule (not all tho). Thanks!

lzap (approved):
All good could you rebase, we have changed CI pipeline. Let's merge.

lzap (approved):
Thanks. Running tests.

PR #103, opened 2021-08-05
Add 'pmsuspended' to possible virtual machine states.

abbbi:
In case a virtual machine goes in to libvirt mode `pmsuspended`,
situation is not handled by fog-libvirt currently. This leads to issues
in other projects like vagrant-libvirt. Suspend mode can be either the
case if domain is set into suspend (by for example systemctl suspend) or
hypervisor sets VM into suspended mode due to lack of resources.

Full discussion:

 vagrant-libvirt/vagrant-libvirt#1200

Patch adds the state to possible virtual machine states.


lzap (approved):
A valid state, yeah, thanks.

PR #104, opened 2021-08-18
README example session

lzap:
An example session for the README


ekohl (approved):
This caused me to look at the README. Looks like we need to clean up the badges as well.

PR #107, opened 2021-10-06
Remove NVRAM in undefine call

NPellet:
1. Use splat to pass any argument to the function call
2. In ```undefine```, pass the argument to remove the NVRAM

Arguably, the flag could come from the caller, and not from fog-libvirt itself. What do you think ?

PR related to https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1372

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -92,7 +92,7 @@ module Fog
 
         def destroy(options={ :destroy_volumes => false})
           poweroff unless stopped?
-          service.vm_action(uuid, :undefine)
+          service.vm_action(uuid, :undefine, 4) # Remove NVRAM if exists
           volumes.each { |vol| vol.destroy } if options[:destroy_volumes]
           true
         end
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -92,7 +92,7 @@ module Fog
 
         def destroy(options={ :destroy_volumes => false})
           poweroff unless stopped?
-          service.vm_action(uuid, :undefine, 4) # Remove NVRAM if exists
+          service.vm_action(uuid, :undefine, Libvirt::Domain::UNDEFINE_NVRAM) # Remove NVRAM if exists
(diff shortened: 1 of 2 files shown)


ekohl:
Could you help me understand what NVRAM is the context of libvirt and why it needs to be undefined in the stop action? What are the implications for machines?

NPellet:
I'm probably not the right person to do that, but my understanding is that UEFI physical machines have some kind of NVRAM chip which stores some state permenantly (like the boot order and whatnot).
OVMF emulates that using an OVMF_VARS.fd file somewhere on the system and which we have to define in /domain/os/nvram.
In libvirt, when using UEFI, one has to assign a unique NVRAM variable file per domain, as far as I understand. 

When the domain is undefined, we need to tell libvirt if we should keep the NVRAM file or not, hence the flag I added. Without it, the undefine call fails.

lzap:
LGTM however the only concern is compatibility. This was added in:

https://github.com/libvirt/libvirt/commit/273b6581ca8dae11e6ff40e3d13813fdbb37d41b

into 1.2.6 version, pretty old it looks like but RHEL 7 hypervisors are still pretty common. So to double check: https://access.redhat.com/downloads/content/libvirt/1.1.1-12.el7/x86_64/f21541eb/package

* RHEL 7 ships with libvirt 1.1 - 4.5
* RHEL 6 ships with librit 0.8.7 - 0.10.2

Therefore I recommend to make a decision to support libvirt 4.5 and higher (EL7 users are required to update to the latest version). Can you make a note in the README that the oldest supported libvirt version is 4.5?

Test fail.

ekohl:
My intuition says it should be opt-in. That would also keep the API compatible. However, I don't know enough about it to answer it with certainty.

lzap:
We are just about to close the 0.9.1 release, please push now if you want this in. An opt-in argument we can accept without fear.

ekohl:
At this point I have fewer concerns for compatibility. By now there's at least a merge conflict with 085a7c89596e989921dccd33dd954afbb37df27a. I've opened https://github.com/fog/fog-libvirt/pull/132, which I think should fix implement this.

PR #109, opened 2022-01-18
Ignore owner/group when not set

lzap:
SSIA

This is untested, asked a customer to test it out for me.

### DIFF
--- a/lib/fog/libvirt/models/compute/templates/volume.xml.erb
+++ b/lib/fog/libvirt/models/compute/templates/volume.xml.erb
@@ -20,8 +20,12 @@
           <path><%= backing_volume.path %></path>
           <format type="<%= backing_volume.format_type %>"/>
           <permissions>
+            <% if owner -%>
             <owner><%= owner %></owner>
+            <% end -%>
+            <% if group -%>
             <group><%= group %></group>
+            <% end -%>
             <mode>0744</mode>
             <label>virt_image_t</label>
           </permissions>


lib/fog/libvirt/models/compute/templates/volume.xml.erb
```diff
@@ -20,8 +20,12 @@
           <path><%= backing_volume.path %></path>
           <format type="<%= backing_volume.format_type %>"/>
           <permissions>
+            <% if owner -%>
             <owner><%= owner %></owner>
+            <% end -%>
+            <% if group -%>
             <group><%= group %></group>
+            <% end -%>
             <mode>0744</mode>
```
ekohl, line 29: Unrelated, but what a weird mode.

ekohl (approved):
I thought I had a dejavu and I already addressed this https://github.com/fog/fog-libvirt/pull/87 but looks like I wasn't thorough enough.

lib/fog/libvirt/models/compute/templates/volume.xml.erb
lzap: Indeed but it looks like this is how libvirt creates it by default. This is an image created by vagrant-libvirt:

```
-rwxr--r--. 1 lzap lzap  1735852032 23. pro 11.44 centos-VAGRANTSLASH-stream8_vagrant_box_image_20210210.0.img
```

Well, unless it also uses fog library which I don't think is the case, nothing in their Gemfile.

lzap:
The customer confirmed it works for them: https://community.theforeman.org/t/using-libvirt-image-based-provisioning-error/26032/8

Looks like an useful patch then!

lib/fog/libvirt/models/compute/templates/volume.xml.erb
ekohl: > Well, unless it also uses fog library which I don't think is the case, nothing in their Gemfile.
That's because it's in the gemspec:
https://github.com/vagrant-libvirt/vagrant-libvirt/blob/b3cf54a946600329d691f81cce32d7c407f8d88e/vagrant-libvirt.gemspec#L27

PR #113, opened 2022-04-21
Avoid querying for all networks by default

electrofelix:
Given the network attribute of the nic is assumed to exist, provide this
as a filter attribute to skip querying libvirt for all networks and
instead request only the one relevant.

This can help avoid a race condition where multiple sets of vagrant
machines are being run on the same host using different networks where
during delete/destroy the networks of one are removed and yet at the
same time the other environment may attempt to access the attributes.

Relates-to: vagrant-libvirt/vagrant-libvirt#1438

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -411,13 +411,11 @@ module Fog
           ip_address = nil
           nic = self.nics.find {|nic| nic.mac==mac}
           if !nic.nil?
-            service.networks.all.each do |net|
-              if net.name == nic.network
-                leases = net.dhcp_leases(mac, 0)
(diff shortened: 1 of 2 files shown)


electrofelix:
@geemus I think I've run into a common issue with how filters are currently implemented for fog-libvirt requests. It appears that the structure of the code prevents use of the same logic for the application of the filter. I'm not sure if there is an easy solution to this, just that it appears to make it quite difficult to work out how to write suitable tests and be confident that the behaviour tested aligns with behaviour that will be seen when working with real requests.

geemus:
@electrofelix thanks for the heads up, I'm not sure off hand, but would be happy to talk through it some more if that would be helpful.

geemus (approved):

Issue #110, opened 2022-02-18
list_domains:domains_volumes only parses <source file=...>


idl0r:
Hi,

https://github.com/fog/fog-libvirt/blob/f2f75f0e493351a4cfecd269bbd3c4bb92a27fa6/lib/fog/libvirt/requests/compute/list_domains.rb#L42

that parses only definitions like:
```
    <disk type='file' device='disk'>
      <driver name='qemu' type='raw' cache='none' io='native'/>
      <source file='/dev/mapper/s093043'/>
      <backingStore/>
      <target dev='vda' bus='virtio'/>
      <alias name='virtio-disk0'/>
      <address type='pci' domain='0x0000' bus='0x00' slot='0x06' function='0x0'/>
    </disk>
```

However, there's also "dev" (type block) which seams to be valid as well?
```
    <disk type='block' device='disk'>
      <driver name='qemu' type='raw' cache='none' io='native'/>
      <source dev='/dev/mapper/s093250'/>
      <backingStore/>
      <target dev='vda' bus='virtio'/>
      <alias name='virtio-disk0'/>
      <address type='pci' domain='0x0000' bus='0x00' slot='0x06' function='0x0'/>
    </disk>
```

The latter isn't covered by the above mentioned request.

volumes/volume_path would be nil then.

idl0r:
Uhm, please disable that bot until someone looked at it?

lestercheung:
Hi @ekohl 👋🏼
Now that this bug has been fixed - do you guys have a plan for a new release and bump the version used by Foreman?
Looking forward to hearing from you and your team.

ekohl:
There's a 0.12.0 release out and it's in Foreman nightly now

ekohl:
This should have been closed when #117 was merged.

debdolph:
Foreman is still using 0.9.0 in develop :/ https://github.com/theforeman/foreman/blob/develop/bundler.d/libvirt.rb

ekohl:
No, that's >= 0.9.0 so no upper bound

PR #117, opened 2022-06-15
Add handling of block volumes with a dev source.

rwf14f:
This fixes issue #110.

### DIFF
--- a/lib/fog/libvirt/requests/compute/list_domains.rb
+++ b/lib/fog/libvirt/requests/compute/list_domains.rb
@@ -39,13 +39,9 @@ module Fog
         end
 
         def domain_volumes xml
-          vols_by_file = xml_elements(xml, "domain/devices/disk/source", "file")
-          vols_by_name = xml_elements(xml, "domain/devices/disk/source", "name")
-          vols = []
-          vols_by_file.zip(vols_by_name).each do |by_file,by_name|
-            vols.push(by_file.nil? ? by_name : by_file)
+          xml_elements(xml, "domain/devices/disk/source").map do |element|
+            element[:file] || element[:dev] || element[:name]
           end
-          vols
         end
 
         def boot_order xml


lib/fog/libvirt/requests/compute/list_domains.rb
```diff
@@ -40,10 +40,15 @@ def domain_display xml
...
           vols_by_name = xml_elements(xml, "domain/devices/disk/source", "name")
           vols = []
-          vols_by_file.zip(vols_by_name).each do |by_file,by_name|
-            vols.push(by_file.nil? ? by_name : by_file)
+          vols_by_file.zip(vols_by_dev,vols_by_name).each do |by_file,by_dev,by_name|
+            if by_file.nil?
+              vols.push(by_dev.nil? ? by_name : by_dev)
+            else
+              vols.push(by_file)
+            end
```
ekohl, line 51: Just thinking out loud. Isn't this equivalent?
```suggestion
            vols.push(by_file || by_dev || by_name)
```

And you could even go a step further and replace `.each` with `.map`, then get rid of the `vols` variable altogether, but that may be too much of a refactor for this patch.

ekohl (commented):
Overall the code looks ok to me. I do wonder if you could read the XML element source and then iterate over that list rather than iterating over the same thing 3 times and then zipping them back together.

lib/fog/libvirt/requests/compute/list_domains.rb
```diff
@@ -39,13 +39,9 @@ def domain_display xml
...
-          vols_by_file = xml_elements(xml, "domain/devices/disk/source", "file")
-          vols_by_name = xml_elements(xml, "domain/devices/disk/source", "name")
-          vols = []
-          vols_by_file.zip(vols_by_name).each do |by_file,by_name|
-            vols.push(by_file.nil? ? by_name : by_file)
-          end
-          vols
+          xml_elements(xml, "domain/devices/disk/source").map{|element|
+            element["file"] || element["dev"] || element["name"]
+          }
```
ekohl, line 44: Sorry for not catching this previously, but it's common to use `do/end` when you have multiple lines.

I don't have a test env at hand, but in the util function they're retrieved via symbols. Should that also be done here as well?

```suggestion
          xml_elements(xml, "domain/devices/disk/source").map do |element|
            element[:file] || element[:dev] || element[:name]
          end
```

rwf14f: > Sorry for not catching this previously, but it's common to use `do/end` when you have multiple lines.

No problem to change this, but then we should change it in ``domain_interfaces`` as well for consistency.

> I don't have a test env at hand, but in the util function they're retrieved via symbols. Should that also be done here as well?

It works with symbols, so no problem changing it, probably also better for consistency. I just used strings because that's used in the Nokogiri rdocs.

lestercheung:
Is there any plans to mere this? libvirtd in EL8 won't work if the configuration refers to a LVM block device using the "file" attribute so would love to see this merged.

This [foreman forum post bought me here](https://community.theforeman.org/t/cannot-edit-libvirt-hosts/27812).

ekohl (approved):
Thanks for addressing this. It looks good to me.

PR #126, opened 2022-12-13
:volumes should return empty array instead of array of nils

stejskalleos:


stejskalleos:
@ekohl

ekohl:
I should add that I tried to see if this was needed in the default setup but https://github.com/ekohl/fog-libvirt/commit/107e673a1147aa4626e4ccca0107e05a39bfc00b suggests this doesn't make a difference. Perhaps we set something up in Foreman's tests. I do wonder if #117 perhaps changed the output format.

ekohl:
Also, it should be noted that Foreman's tests in CI no longer fail with 0.11.0 so is this still an issue? Perhaps it was triggered by the failed delete?

stejskalleos:
I don't know why but locally `test/controllers/compute_resources_vms_controller_test.rb` is still failing for me, even on fresh DB
```
ComputeResourcesVmsControllerTest#test_0008_should show vm:
ActionView::Template::Error: undefined method `allocation' for nil:NilClass
    app/views/compute_resources_vms/show/_libvirt.html.erb:46
```

But I tried to downgrade fog-libvirt and issue is still there, so I guess it's problem on (my) Foreman. Closing the PR for now

PR #127, opened 2023-07-03
Set the default machine type to q35 on i686 and x86_64

ekohl:
The pc type has been deprecated and on EL9 it's even unsupported. The type on aarch64 is virt, so there we continue to rely on the default.

This PR includes https://github.com/fog/fog-libvirt/pull/94 because it's easier to build on than the current template.

Right now it's untested whether this XML actually works.

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -295,12 +295,16 @@ module Fog
               end
 
               unless cpu.empty?
-                if cpu[:mode]
-                  xml.cpu(:mode => cpu[:mode])
-                else
+                if cpu.dig(:model, :name)
                   xml.cpu do
                     xml.model(cpu.dig(:model, :name), :fallback => cpu.dig(:model, :fallback) || "allow")
                   end
+                else
+                  xml.cpu(
+                    :mode => cpu.dig(:model, :name) || "host-passthrough",
+                    :check => cpu.fetch(:check, "none"),
+                    :migratable => cpu.fetch(:migratable, "on")
+                  )
                 end
               end
 
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -282,7 +282,8 @@ module Fog
 
               xml.vcpu(cpus)
               xml.os do
-                xml.type(os_type, :arch => arch)
+                type = xml.type(os_type, :arch => arch)
+                type[:machine] = "q35" if ["i686", "x86_64"].include?(arch)
 
                 boot_order.each do |dev|
                   xml.boot(:dev => dev)
@@ -291,7 +292,6 @@ module Fog
               xml.features do
                 xml.acpi
                 xml.apic
-                xml.pae
               end
 
               unless cpu.empty?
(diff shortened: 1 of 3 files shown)


ekohl:
@jhoblitt did you verify this patch?

jhoblitt:
I am testing with Almalinux 9.2 as the libvirt host.  The generated XML causes almalinux 9.2 to panic while booting vmlinuz.  I tried centos stream 9.2 as well but get an ipxe error I don't understand.... it might be a tls error... 

I generated a "reference" VM using `virt-install` in order to compare the XMl against. 

```bash
# virt-install --name=jhoblitt-test9 --vcpus=4 --ram=8192 --file-size=40 --machine=q35 --network bridge=br2101 --location=/tmp/AlmaLinux-9.2-x86_64-minimal.iso
```

I'm working on minimum needed changes to get almalinux 9.2 to boot.

jhoblitt:
This was enough to allow EL7.9, EL8.8, and EL9.2 to install on libvirt from EL9.2:

```
diff --git a/lib/fog/libvirt/models/compute/server.rb b/lib/fog/libvirt/models/compute/server.rb
index c2c8f47..eafad59 100644
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -292,20 +292,21 @@ module Fog
               xml.features do
                 xml.acpi
                 xml.apic
-                xml.pae
               end
 
               unless cpu.empty?
                 if cpu[:mode]
                   xml.cpu(:mode => cpu[:mode])
                 else
-                  xml.cpu do
-                    xml.model(cpu.dig(:model, :name), :fallback => cpu.dig(:model, :fallback) || "allow")
-                  end
+                  xml.cpu(:mode => 'host-passthrough', :check=>'none', :migratable => 'on')
                 end
               end
 
-              xml.clock(:offset => "utc")
+              xml.clock(:offset => "utc") do
+                xml.timer(:name => 'rtc', :tickpolicy=> 'catchup')
+                xml.timer(:name => 'pit', :tickpolicy=> 'delay')
+                xml.timer(:name => 'hpet', :present=> 'no')
+              end
 
               xml.devices do
                 ceph_args = read_ceph_args
```

ekohl:
Thanks for that.

I guess PAE doesn't matter anymore because x86-64 made it obsolete and I doubt anyone wants to do virtualization on i686.

I do wonder about the CPU model, because that would ignore some parameters. And a host-passthrough mode and migratable would only work if you have the same CPU model in both machines. Now I suspect fog-libvirt would rarely be used in such a context, so it probably doesn't make a big difference.

I'd need to look into the timers documentation, but it's also what virt-manager did for my Debian machine so I'd feel confident in it.

jhoblitt:
Changing only the pae flag and the clock wasn't enough to stop the kernel panic.  I did not test pae + cpu mode without the clock config changes.  I highly doubt that pass-through mode is required but some sort of change to the cpu mode was definitely needed.

jhoblitt:
@ekohl Could I do anything to help this along?

ekohl:
I do see that virt-manager also prefers host pass through by default. I've chosen a slightly different way because I think it still allows overriding.

Note I'll be on vacation starting Friday and I have to do Foreman 3.8 branching preparations before then so I'm not sure I can come back to this before I come back.

jhoblitt:
@ekohl What is remaining before this can be merged?

jhoblitt (approved):
This PR has been deployed into my production env and is known to be working.

https://github.com/lsst-it/lsst-control/blob/production/site/profile/manifests/core/foreman/fog_hack.pp

jhoblitt:
Ping?

ekohl:
Sorry, so many things going on. I'll see about releasing a fog-libvirt 1.0.0 with these changes included.

PR #129, opened 2023-07-17
Fix disk type for block devices on QEMU >= 6.0

ekohl:
Since [QEMU version 6.0](https://wiki.qemu.org/ChangeLog/6.0#Incompatible_changes), you can't use the file block driver with block devices. It tries to detect block devices by looking at the file path.

This builds on https://github.com/fog/fog-libvirt/pull/116, but I don't have the capacity to test this myself right now.

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -331,9 +331,14 @@ module Fog
                       xml.target(:dev => target_device, :bus => args["bus_type"] == "virtio" ? "virtio" : "scsi")
                     end
                   else
-                    xml.disk(:type => "file", :device => "disk") do
+                    is_block = volume.path.start_with?("/dev/")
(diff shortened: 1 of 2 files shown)


ekohl:
@lero @rwf14f  @antaflos since you were involved in #116, could you have a look?

antaflos:
Ok, seems like the vendored fog-libvirt gem in Foreman is on version 0.11.0, which is quite different from the current master branch (using XML builder instead of an ERB template), so applying this patch manually won't work.

rwf14f:
The patch looks good. I've tested it on our Foreman server:
- replaced fog-libvirt-0.11 with github master branch
- applied patch
- restarted Foreman
- deployed a VM using an LVM pool (tested qemu 6.2 and 7.2)
- deployed a VM using a directory pool (qemu 7.2)

In those cases, it produced a valid XML description and the VMs started without problems.

ekohl:
Thanks for confirming @rwf14f!

PR #133, opened 2023-10-20
Fix possible nil volumes for libvirt

ofedoren:
@ekohl, whist I was trying to run Foreman tests on Ruby 2.7/3.0 via GA I've noticed that sometimes `volumes` can be `[nil]`. It's probably an issue with the setup, but the fix still makes this safer.


ekohl (approved):
Ruby 2.7+ has `.filter_map` for `.map.compact` but this gem is currently aiming for 2.5+ so this is correct.

PR #132, opened 2023-10-19
Undefine NVRAM on vm destruction

ekohl:
VMs with UEFI store UEFI variables in OVMF files. This flag tells Libvirt to undefine it, if it exists. This was introduced in libvirt 1.2.6, which can be found on very old EL7 systems, but those are no longer supported. EL 7.9 (only supported EL7 version) ships libvirt 4.5.

This replaces https://github.com/fog/fog-libvirt/pull/107

PR #134, opened 2023-12-20
Implement EFI host creation

### LINKED ISSUE #128: No ability to create libvirt KVM virtual machine instances with uefi support
Currently fog-libvirt only supports ability to create virtual machines via libvirt that are BIOS based. BIOS is slowly being deprecated on with Linux distributions. While Windows 11 no longer supports it all together.

Foreman currently relies on fog-libvirt to create virtual machines from a libvirt compute resource. Currently it can only support creation of BIOS based KVM virtual machines. 

Enabling a way to specify creation of libvirt KVM instances with UEFI will resolve this issue.

https://community.theforeman.org/t/libvirt-and-uefi/22756

ekohl:
This implements EFI host creation by specifying the os_firmware attribute to be efi. It's also possible to set the os_loader attribute to secure (for UEFI secure boot) or stateless (for AMD SEV). Both secure and stateless imply the os_firmware to be efi.

It includes #132 because the secure boot loader creates NVRAM and otherwise you wouldn't be able to delete those machines.

Only minimal testing on Fedora 38 has been done, but I can create VMs and have them boot up.

Fixes #128

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -13,7 +13,9 @@ module Fog
 
         attribute :cpus
         attribute :cputime
+        attribute :os_firmware
         attribute :os_type
+        attribute :os_loader
         attribute :memory_size
         attribute :max_memory_size
         attribute :name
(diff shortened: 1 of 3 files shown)


ekohl:
I really don't like the RuboCop style rules that are set up ...

jhoblitt:
What is needed to move forward with this?  I would like to start using efi for pxebooting libvirt VMs via foreman.

lib/fog/libvirt/models/compute/server.rb
```diff
@@ -281,10 +283,28 @@ def to_xml
...
+              elsif os_firmware
+                os_tags[:firmware] = os_firmware
+              end
+
+              xml.os(**os_tags) do
                 type = xml.type(os_type, :arch => arch)
                 type[:machine] = "q35" if ["i686", "x86_64"].include?(arch)
 
+                # TODO: can you use both secure and stateless at the same time?
+                case attributes[:os_loader]
```
ekohl, line 301: Note to self: https://libvirt.org/kbase/secureboot.html says the os_loader was only needed for libvirt >= 7.2 < 8.6

ekohl:
Made redundant by #155.

PR #135, opened 2023-12-21
Always use XML builders to construct XML

ekohl:
This removes the use of ERB to construct XML and instead always uses a builder.

The primary motivation was that the ERB invocation was raising an error and an XML builder guarantees valid XML. While doing this I figured out that at least Pool doesn't have a .to_xml and Network.to_xml was just broken. Pool.to_xml does work, but it is noted in the docs (https://libvirt.org/formatstorage.html) that some fileds (`allocation`, `capacity` & `available`) are not applicable when creating a pool so it's questionable if they should be included.

Given that, I question if anyone actually uses this in practice, so I'm not too worried about breaking things.


lib/fog/libvirt/models/compute/network.rb
```diff
@@ -27,6 +27,17 @@ def save
         def shutdown
           service.destroy_network(uuid)
         end
+
+        def to_xml
```
ekohl, line 31: Further testing told me `save` isn't even implemented now so I'm not too worried about this.

Issue #136, opened 2024-03-18
Q35 machine type does not support IDE


bmagistro:
When trying to use a cloud-init image based deployment with Foreman, the VM provisioning fails as the Q35 machine type does not support the IDE bus.  This was broken by #127 a change similar to this fix was mentioned as part of #76  but does not look like it made it into the final set.  I haven't been able to figure out the conditional check yet so do not have a pr but the beloow changes have been confirmed to work under q35.  I have not looked to see if `i440` supports sata allowing the logic to be skipped/deferred for now.  If you have suggestions to query the machine, I can try to finish turning this into a pr.

```
xml.target(:dev => "hdc", :bus => "ide") --> xml.target(:dev => "sr0", :bus => "sata")
xml.address(:type => "drive", :controller => 0, :bus => 1, :unit => 0) --> xml.address(:type => "drive", :controller => 0, :bus => 0, :unit => 0)
```

This issue is mentioned in serval places
* https://bugs.launchpad.net/nova/+bug/1831538
* https://github.com/dmacvicar/terraform-provider-libvirt/issues/885
* https://bugzilla.redhat.com/show_bug.cgi?id=1707351

smirta:
Hi @bmagistro, we ran into the same issue. I added a PR #137 but chose SCSI instead of SATA. Cheers

PR #137, opened 2024-06-02
feat: replaced unsupported ide with q35 machine type with scsi cdrom …

smirta:
Implements SCSI as cdrom driver instead of IDE because it is incompatible with machine type q35 as introduced in #127 . It should resolve the issue raised in #136 .


ekohl (commented):
I think the test failures are related. Have you had a look at them?

ekohl (approved):
I took the liberty of doing a trivial rebase after I did some fixes. When I locally ran this it now passes, but I'll also note that we don't have test coverage for this.

smirta:
nice, thank you @ekohl . Unfortunately, I didn't find time to sort out the issue. Glad you did :)

PR #139, opened 2024-06-12
Correct the VNC password attribute

ekohl:
In the refactor to the XML builder a subtle bug was introduced which caused the password not to be set. This uses the correct attribute.

Came up in https://community.theforeman.org/t/error-creating-libvirt-based-guest-after-upgrading-to-3-10/38295. I'm asking the reporter to test this patch.

Fixes: 6db6f8089418 ("Replace server XML with Nokogiri::XML::Builder")

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -408,7 +408,7 @@ module Fog
                     graphics.autoport = "yes"
                   end
                   graphics.listen = display[:listen] unless display[:listen].empty?
-                  graphics.password = display[:password] unless display[:password].empty?
+                  graphics.passwd = display[:password] unless display[:password].empty?
 
                   xml.video do
                     xml.model(:type => "cirrus", :vram => 9216, :heads => 1)


jose-deniz-ns90:
Patch has been tested successfully on fog-libvirt (from package rubygem-fog-libvirt-0.12.0-1.el8.noarch)
Thanks!

ekohl:
I see the failure also happens in master, so I wonder what changed.

ekohl:
Looks like https://gitlab.com/libvirt/libvirt-ruby/-/commit/27907495e7ecf8122b0ea3e66e3fa34d2714b5a8 introduced this regression.

ekohl:
I started the fixes in #140.

lestercheung:
Confirmed working on a v3.9.1 katello installation 🙏🏼

ekohl:
In https://projects.theforeman.org/issues/37352 it was brought up that if the VM is created incorrectly (without a password) that it can't be change later. So the `set_password` fails.

I'm not sure yet what needs to happen to rectify this.

PR #140, opened 2024-06-13
Use test:///default in tests

ekohl:
Rather than mocking, this uses the test connection to use real libvirt objects.

It only keeps mocking for dhcp_leases which isn't implemented in the test connection. The UUIDs are modified to match.

This was suggested by the ruby-libvirt maintainers in https://gitlab.com/libvirt/libvirt-ruby/-/merge_requests/35#note_1948893893. Right now it still contains various unrelated patches, though some are certainly needed. I'd qualify some of them as features, others bug fixes and some general cleanups.


ekohl:
I think this is ready now.

I'm tempted to create a bugfix release with #142 and #139 first with the dependency pinned and then merge this.

voxik:
@ekohl is there by a chance any ETA getting this out of the doors?

ekohl:
I was hoping we'd finish the UEFI and secureboot things, but I can release what's already out there. Probably won't be this week though.

PR #143, opened 2024-06-24
minitest-stub-const is not use anywhere

voxik:
According to log, it wasn't ever used.

### DIFF
--- a/fog-libvirt.gemspec
+++ b/fog-libvirt.gemspec
@@ -33,7 +33,6 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency("net-ssh")
   s.add_development_dependency("minitest", "~> 5.0")
-  s.add_development_dependency("minitest-stub-const")
   s.add_development_dependency("pry")
   s.add_development_dependency("rake")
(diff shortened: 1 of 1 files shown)


voxik:
The test matrix would probably deserve some update 🤔

ekohl:
Yes, and overall #140 is one we need to finish because currently all tests are failing. Though I've been thinking about pinning the dependency in the short term.

voxik:
> because currently all tests are failing. Though I've been thinking about pinning the dependency in the short term.

Yes, ruby-libvirt 0.8.2 should fix the test failures. I think this is the issue:

~~~
$ ruby -r libvirt -e '::Libvirt::Domain.new()'
-e:1:in `new': allocator undefined for Libvirt::Domain (TypeError)
        from -e:1:in `<main>'
~~~

voxik:
Ah, see your discussion in https://gitlab.com/libvirt/libvirt-ruby/-/merge_requests/35 now

ekohl:
I've pinned it and taken the liberty to rebase your PR.

PR #146, opened 2024-07-12
Correctly pass UNDEFINE_NVRAM when flags isn't pased

### LINKED ISSUE #142: Undefine NVRAM not working if options without flags are passed to destroy
The code introduced with https://github.com/fog/fog-libvirt/pull/132 doesn't work when an `options` hash is passed to `destroy` that doesn't define `:flags`.
For example, if `{ :destroy_volumes => true }` is passed to `destroy`, then the if condition is false because it uses the default value of the `options.fetch` and then passes an undefined `options[:flags]` to the `vm_action` call in the else block.
If the intention behind the patch is to use `UNDEFINE_NVRAM` as the default if no flags are passed to `destroy` then this should work:
```
flags = options.fetch(:flags, ::Libvirt::Domain::UNDEFINE_NVRAM)
if flags.zero?
  service.vm_action(uuid, :undefine)
else
  service.vm_action(uuid, :undefine, flags)
end
```

ekohl:
If `destroy()` is called with options that doesn't contain a `flags` value then it should still default to `UNDEFINE_NVRAM`. That wasn't happening.

Fixes: 21d1bf827c03 ("Undefine NVRAM on vm destruction")

Fixes #142

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -92,10 +92,11 @@ module Fog
 
         def destroy(options={ :destroy_volumes => false, :flags => ::Libvirt::Domain::UNDEFINE_NVRAM })
           poweroff unless stopped?
-          if options.fetch(:flags, ::Libvirt::Domain::UNDEFINE_NVRAM).zero?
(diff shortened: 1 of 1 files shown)


ekohl:
I learned that the `test://default` driver doesn't implement this flag. #153 tracks it further.

PR #153, opened 2024-07-16
Filter out UNDEFINE_NVRAM for test:///default

ekohl:
The test driver doesn't support UNDEFINE_NVRAM and I couldn't find a way to probe for the capability. This hacks around it by looking at the service uri and filters out the flag.

It also adds an explicit test since previously the exception was swallowed. A new server is created to avoid interference with other tests.

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -97,6 +97,10 @@ module Fog
           if flags.zero?
             service.vm_action(uuid, :undefine)
           else
+            # the test driver doesn't support UNDEFINE_NVRAM
+            if service.uri.driver == 'test'
+              flags ^= ::Libvirt::Domain::UNDEFINE_NVRAM
+            end
             service.vm_action(uuid, :undefine, flags)
           end
           volumes.each { |vol| vol.destroy } if options[:destroy_volumes]
--- a/tests/libvirt/models/compute/server_tests.rb
+++ b/tests/libvirt/models/compute/server_tests.rb
@@ -13,7 +13,7 @@ Shindo.tests('Fog::Compute[:libvirt] | server model', ['libvirt']) do
       %w{ start stop destroy reboot suspend }.each do |action|
         test(action) { server.respond_to? action }
       end
-      %w{ start reboot suspend stop destroy}.each do |action|
+      %w{ start reboot suspend stop }.each do |action|
         test("#{action} returns successfully") {
           begin
             server.send(action.to_sym)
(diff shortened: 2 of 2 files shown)


ekohl:
https://gitlab.com/libvirt/libvirt/-/blob/239669049d9904e5e8da2d8b2a38d4d927a167e9/src/qemu/qemu_capabilities.c#L261 suggests there's an nvram capability, but I don't see this on my real libvirt instance.

https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainUndefineFlagsValues is a list of possible flags, but this really doesn't help since I don't see a way to probe for supported flags.

lib/fog/libvirt/models/compute/server.rb
```diff
@@ -97,6 +97,11 @@ def destroy(options={ :destroy_volumes => false, :flags => ::Libvirt::Domain::UN
           if flags.zero?
             service.vm_action(uuid, :undefine)
           else
+            # the test driver doesn't support UNDEFINE_NVRAM and we can't probe
```
sbernhard, line 100: Should we have a "FIXME" comment so that it gets highlighted?
Is it possible to open a issue for the test driver so that the UNDEFINE_NVRAM gets fixed soonish?

lib/fog/libvirt/models/compute/server.rb
```diff
@@ -97,6 +97,11 @@ def destroy(options={ :destroy_volumes => false, :flags => ::Libvirt::Domain::UN
...
+            # for the capability at runtime
+            if service.uri.uri == 'test:///default'
```
ekohl, line 102: I did some further digging. https://libvirt.org/drvtest.html documents the test driver. So really it should look at the scheme see if that's `test` or `test+something`.

lib/fog/libvirt/models/compute/server.rb
ekohl: Some more digging in the source tells me it's implemented in [src/test](https://gitlab.com/libvirt/libvirt/-/tree/master/src/test?ref_type=heads). Specifically, https://gitlab.com/libvirt/libvirt/-/blob/239669049d9904e5e8da2d8b2a38d4d927a167e9/src/test/test_driver.c?page=5#L4578-4634

I'm not sure if it can simply be added to `virCheckFlags` there, or if the whole driver needs to support nvram.

sbernhard: this is true. The https://gitlab.com/libvirt/libvirt/-/blob/master/src/internal.h?ref_type=heads#L259 is a just a macro and will return if a unsupported flag is used. And https://gitlab.com/libvirt/libvirt/-/blob/239669049d9904e5e8da2d8b2a38d4d927a167e9/src/test/test_driver.c?page=5#L4587 only specified some flags but not NVRAM.

dosas (approved):
LGTM. Thanks for fixing this.

PR #155, opened 2024-07-22
EFI & Secure Boot

nofaralfasi:
EFI and Secure Boot support for Libvirt. 

Required for: https://github.com/theforeman/foreman/pull/10321

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -13,8 +13,10 @@ module Fog
 
         attribute :cpus
         attribute :cputime
-        attribute :os_firmware
-        attribute :os_firmware_features
+        attribute :firmware
+        attribute :firmware_features
+        attribute :secure_boot
+        attribute :loader_attributes
         attribute :os_type
         attribute :memory_size
         attribute :max_memory_size
@@ -291,8 +293,7 @@ module Fog
               xml.vcpu(cpus)
               os_tags = {}
 
-              # Set firmware only if it's EFI, BIOS don't need to be set
-              os_tags[:firmware] = "efi" if os_firmware == "efi"
+              os_tags[:firmware] = firmware if firmware == 'efi'
 
               xml.os(**os_tags) do
                 type = xml.type(os_type, :arch => arch)
@@ -302,9 +303,13 @@ module Fog
                   xml.boot(:dev => dev)
                 end
 
-                if os_firmware == "efi"
+                loader_attributes&.each do |key, value|
+                  xml.loader(key => value)
+                end
+
+                if firmware == "efi" && firmware_features&.any?
                   xml.firmware do
-                    os_firmware_features.each_pair do |key, value|
+                    firmware_features.each_pair do |key, value|
                       xml.feature(:name => key, :enabled => value)
                     end
                   end
@@ -555,6 +560,7 @@ module Fog
             :guest_agent            => true,
             :video                  => {:type => "cirrus", :vram => 9216, :heads => 1},
             :virtio_rng             => {},
+            :firmware_features      => { "secure-boot" => "no" },
           }
         end
 
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -13,6 +13,8 @@ module Fog
 
         attribute :cpus
         attribute :cputime
+        attribute :os_firmware
+        attribute :os_firmware_features
         attribute :os_type
         attribute :memory_size
         attribute :max_memory_size
@@ -287,14 +289,28 @@ module Fog
               end
 
               xml.vcpu(cpus)
-              xml.os do
+              os_tags = {}
+
+              # Set firmware only if it's EFI, BIOS don't need to be set
+              os_tags[:firmware] = "efi" if os_firmware == "efi"
+
+              xml.os(**os_tags) do
                 type = xml.type(os_type, :arch => arch)
                 type[:machine] = "q35" if ["i686", "x86_64"].include?(arch)
 
                 boot_order.each do |dev|
                   xml.boot(:dev => dev)
                 end
+
+                if os_firmware == "efi"
+                  xml.firmware do
+                    os_firmware_features.each_pair do |key, value|
+                      xml.feature(:name => key, :enabled => value)
+                    end
+                  end
+                end
               end
+
               xml.features do
                 xml.acpi
                 xml.apic
--- a/lib/fog/libvirt/requests/compute/list_domains.rb
+++ b/lib/fog/libvirt/requests/compute/list_domains.rb
@@ -46,6 +46,24 @@ module Fog
           xml_elements(xml, "domain/os/boot", "dev")
         end
 
+        def firmware(xml)
+          firmware_from_loader = xml_elements(xml, "domain/os/loader", "type").first
+
+          case firmware_from_loader
+          when 'pflash'
+            'efi'
+          when 'rom'
+            'bios'
+          else
+            xml_elements(xml, "domain/os", "firmware").first || 'bios'
+          end
+        end
+
+        # we rely on the fact that the secure attribute is only present when secure boot is enabled
+        def secure_boot_enabled?(xml)
+          xml_elements(xml, "domain/os/loader", "secure").first == 'yes'
+        end
+
         def domain_interfaces xml
           ifs = xml_elements(xml, "domain/devices/interface")
           ifs.map { |i|
@@ -78,7 +96,9 @@ module Fog
               :boot_order      => boot_order(dom.xml_desc),
               :nics            => domain_interfaces(dom.xml_desc),
               :volumes_path    => domain_volumes(dom.xml_desc),
-              :state           => states[dom.info.state]
+              :state           => states[dom.info.state],
+              :firmware        => firmware(dom.xml_desc),
+              :secure_boot     => secure_boot_enabled?(dom.xml_desc),
             }
           rescue ::Libvirt::RetrieveError, ::Libvirt::Error
             # Catch libvirt exceptions to avoid race conditions involving
--- a/tests/libvirt/models/compute/server_tests.rb
+++ b/tests/libvirt/models/compute/server_tests.rb
@@ -32,6 +32,8 @@ Shindo.tests('Fog::Compute[:libvirt] | server model', ['libvirt']) do
       attributes = [ :id,
         :cpus,
         :cputime,
+        :os_firmware,
+        :os_firmware_features,
         :os_type,
         :memory_size,
         :max_memory_size,
@@ -67,6 +69,7 @@ Shindo.tests('Fog::Compute[:libvirt] | server model', ['libvirt']) do
 
     test('be a kind of Fog::Libvirt::Compute::Server') { server.kind_of? Fog::Libvirt::Compute::Server }
     tests("serializes to xml") do
+      test("without firmware") { server.to_xml.include?("<os>") }
       test("with memory") { server.to_xml.match?(%r{<memory>\d+</memory>}) }
       test("with disk of type file") do
         xml = server.to_xml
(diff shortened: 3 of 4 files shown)


nofaralfasi:
I encountered an issue with the changes in this PR. For VMs not created by Foreman (either manually or by another tool), the firmware type is not always set, making it unclear what the firmware type is.
This affects the display of the correct firmware for these hosts in Foreman.

tests/libvirt/models/compute/server_tests.rb
```diff
@@ -79,5 +82,45 @@
...
+          },
+          :nics => [],
+          :volumes => []
+        }
+      )
+      xml = server.to_xml
+
+      os_firmware = xml.include?('<os firmware="efi">')
+      secure_boot = !xml.include?('<feature name="secure-boot" enabled="no" />')
+      enrolled_keys = !xml.include?('<feature name="enrolled-keys" enabled="no" />')
```
ekohl, line 101: I would expect these features to show up.
```suggestion
      secure_boot = xml.include?('<feature name="secure-boot" enabled="no" />')
      enrolled_keys = xml.include?('<feature name="enrolled-keys" enabled="no" />')
```

If they're not expected, I'd use a much broader matcher just to make sure it's not some XML formatting:
```suggestion
      secure_boot = !xml.include?('secure-boot')
      enrolled_keys = !xml.include?('enrolled-keys')
```

nofaralfasi: Yeah, my bad. I forgot to fix this.

nofaralfasi:
> Please look at my `os_loader` implementation around https://github.com/fog/fog-libvirt/pull/134/files#r1686901783. I'm not sure why @stejskalleos dropped the `os_loader`, but from reading https://libvirt.org/kbase/secureboot.html and seeing libvirt 8.0.0 is in EL8 then I suspect you can't use EL8 libvirt & secureboot this way.

You are right; their documentation can be confusing. I was reading from https://libvirt.org/formatdomain.html#bios-bootloader:
`Attribute secure can be used to tell the hypervisor that the firmware is capable of Secure Boot feature. It cannot be used to enable or disable the feature itself in the firmware.`

But I guess that's not true for older versions.

ekohl:
I agree it's often confusing. I'd recommend setting up an EL8 hypervisor (Foreman can connect to libvirt using SSH) and try it out. Often the best way

lib/fog/libvirt/requests/compute/list_domains.rb
```diff
@@ -48,6 +48,23 @@ def boot_order xml
...
+
+          firmware_type
+        end
+
+        def secure_boot_enabled?(xml)
+          enabled_features = xml_elements(xml, "domain/os/firmware/feature")
+                               .select { |feature| feature[:enabled] == 'yes' }
+                               .map { |feature| feature[:name] }
+
+          required_features = ['secure-boot', 'enrolled-keys']
```
ekohl, line 64: AFAIK you don't need `enrolled-keys` for secure-boot. That's just the default set of keys, but users can define their own keys and sign their own kernels.

lib/fog/libvirt/requests/compute/list_domains.rb
```diff
@@ -48,6 +48,23 @@ def boot_order xml
...
+        def firmware(xml)
+          firmware_type = xml_elements(xml, "domain/os", "firmware").first || 'bios'
+          return 'uefi_sb' if firmware_type == 'efi' && secure_boot_enabled?(xml)
+
+          firmware_type
+        end
+
+        def secure_boot_enabled?(xml)
+          enabled_features = xml_elements(xml, "domain/os/firmware/feature")
+                               .select { |feature| feature[:enabled] == 'yes' }
```
ekohl, line 61: Can't you filter it directly with xpath?
```suggestion
          enabled_features = xml_elements(xml, "domain/os/firmware/feature[@enabled='yes']")
```

lib/fog/libvirt/requests/compute/list_domains.rb
```diff
@@ -48,6 +48,23 @@ def boot_order xml
           xml_elements(xml, "domain/os/boot", "dev")
         end
 
+        # Foreman expects the firmware to be 'uefi_sb' if SB is enabled
+        def firmware(xml)
+          firmware_type = xml_elements(xml, "domain/os", "firmware").first || 'bios'
+          return 'uefi_sb' if firmware_type == 'efi' && secure_boot_enabled?(xml)
+
+          firmware_type
+        end
```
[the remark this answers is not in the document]
ekohl: I think `firmware` should exactly output what was also input as an attribute. So I'd just expect `xml_elements(xml, "domain/os", "firmware")`. Then you also read `firmware_features` back and set that attribute. That keeps it consistent that when you create a VM with certain attributes, you can read them back exactly as they were.

In other words, there's a `to_xml` and this can be seen as a `from_xml`.

lib/fog/libvirt/requests/compute/list_domains.rb
ekohl: https://libvirt.org/kbase/secureboot.html#additional-information says that without `enrolled-keys` it allows running unsigned code, but I wonder if that's really true. This last bit clarifies with how I think it's supposed to work:

> The main difference between using a build of EDKII that has Secure Boot support but without keys enrolled and one that doesn't have Secure Boot support at all is that, with the former, you could enroll your own keys and securely run an operating system that you've built and signed yourself. If you are only planning to run existing, off-the-shelf operating system images, then the two configurations are functionally equivalent.

So the idea is that `enrolled-keys` contains the default keys, but if you don't have or want those you can enroll your own keys.

lib/fog/libvirt/requests/compute/list_domains.rb
```diff
@@ -48,6 +48,23 @@ def boot_order xml
...
+
+        def secure_boot_enabled?(xml)
```
ekohl, line 59: Another thought I had: you can implement `secure_boot?` on the `Server` object that inspects the attributes.

nofaralfasi:
I have another idea on how to implement this. Moving it to WIP until it's ready.

lib/fog/libvirt/models/compute/server.rb
```diff
@@ -13,6 +13,10 @@ class Server < Fog::Compute::Server
 
         attribute :cpus
         attribute :cputime
+        attribute :firmware
+        attribute :firmware_features
+        attribute :secure_boot
+        attribute :loader
```
stejskalleos, line 19: Note to myself:
The `loader` element points to a specific boot binary (BIOS/UEFI), while `firmware` refers to the type or particular features used to configure or load the guest VM.

stejskalleos (approved):
:green_apple: **LGTM**

Tested together with https://github.com/theforeman/foreman/pull/10321, BIOS, UEFI & Secure Boot works fine.

lib/fog/libvirt/requests/compute/list_domains.rb
nofaralfasi: For what use?

stejskalleos:
@ekohl, can we get this so the https://github.com/theforeman/foreman/pull/10321 can be included in the next Foreman's  release?

lib/fog/libvirt/models/compute/server.rb
ekohl: I think those could be good comments around the attributes.

To me `loader` implies it's singular, but when I look at the code I see it's a list, so should it be
```suggestion
        attribute :loaders
```

lib/fog/libvirt/requests/compute/list_domains.rb
```diff
@@ -46,6 +46,24 @@ def boot_order xml
...
+          when 'rom'
+            'bios'
+          else
+            xml_elements(xml, "domain/os", "firmware").first || 'bios'
+          end
+        end
+
+        # we rely on the fact that the secure attribute is only present when secure boot is enabled
+        def secure_boot_enabled?(xml)
+          xml_elements(xml, "domain/os/loader", "secure").first == 'yes'
```
ekohl, line 64: Untested, but should this be:
```suggestion
          xml_element(xml, "domain/os/loader", "secure") == 'yes'
```

lib/fog/libvirt/requests/compute/list_domains.rb
```diff
@@ -46,6 +46,24 @@ def boot_order xml
...
+        def firmware(xml)
+          firmware_from_loader = xml_elements(xml, "domain/os/loader", "type").first
+
+          case firmware_from_loader
+          when 'pflash'
+            'efi'
+          when 'rom'
+            'bios'
+          else
+            xml_elements(xml, "domain/os", "firmware").first || 'bios'
```
ekohl, line 58: Untested, but should this be
```suggestion
            xml_element(xml, "domain/os", "firmware") || 'bios'
```

lib/fog/libvirt/requests/compute/list_domains.rb
```diff
@@ -46,6 +46,24 @@ def boot_order xml
           xml_elements(xml, "domain/os/boot", "dev")
         end
 
+        def firmware(xml)
+          firmware_from_loader = xml_elements(xml, "domain/os/loader", "type").first
```
ekohl, line 50: Untested, but should this be
```suggestion
          firmware_from_loader = xml_element(xml, "domain/os/loader", "type")
```

ekohl (commented):
Overall this looks good to me. Some minor implementation notes

lib/fog/libvirt/requests/compute/list_domains.rb
nofaralfasi: `xml_elements(xml, "domain/os/loader", "secure")` returns an array, and the same applies to all other suggestions as well.

ekohl: But that's why I'm suggesting `xml_element` (singular). That code already has a `.first`:
https://github.com/fog/fog-libvirt/blob/11f3821777becf14b3bb3c14122f8cb5af55a8c7/lib/fog/libvirt/models/compute/util/util.rb#L9-L12

nofaralfasi: Oh sorry, I missed that change. Checking.

nofaralfasi: The problem is that in some cases the xml element (`xml/path`) is empty, resulting in an empty array. In such cases, if we use `xml_element` we'd get an error for trying to call `.first[attribute]` on nil. 
To address this, we can either modify the `xml_element` method to handle this issue or continue using `xml_elements` instead.

ekohl: I was afraid of that because I already noticed it when looking at the method. Short term I'd lean to accepting `xml_elements`. Changing `xml_element` to silently return `nil` if the element doesn't exist could break existing code (because it wants to distinguish between the element not existing vs the attribute not existing) and I'd rather not block on figuring that out.

Having said that, I wonder why the code is parsing the XML all the time. Note every `xml_element` and `xml_elements` call parses the XML which probably makes it much slower than it needs to be.

Not a blocker to merging this, but hints for a follow up. I think if `domain_to_attributes` parses the XML and then passes on the parsed tree this would also work:
```suggestion
          loader = (xml / "domain/os/loader")&.first
          loader && loader[:secure] == 'yes'
```

That means you only parse the XML once instead of 6 or so times. Then I think the code is also simple enough to not need `xml_element` or `xml_elements`.

nofaralfasi: I agree that we can improve this, but I'd suggest opening a separate PR for it since the change also affects existing code, not just the part I added. To implement your suggestion, we'd still need to parse the XML, but we could reduce the number of times we do it.

Would it be acceptable to merge this PR as is and then open a new PR to handle the refactoring?

ekohl: We can resolve the XML parsing later, but I'd still like to see an answer to https://github.com/fog/fog-libvirt/pull/155#discussion_r1825764058.

lib/fog/libvirt/models/compute/server.rb
nofaralfasi: The loader is not a list, it's a single tag with multiple attributes.

nofaralfasi: And since we're using the same terminology as Libvirt, I believe there's no need to define "firmware" or "loader" as they should already be familiar to users.

ekohl: But in the code I see it's iterated over which made me think it was a list.
```ruby
loader&.each do |key, value|
  xml.loader(key => value)
end
```

Technically I guess it's a hash, but `loader` still made me expect a string.

nofaralfasi: It's a hash, e.g., `loader: { "secure" => "yes" }`.

ekohl (approved):

PR #156, opened 2024-08-26
Allow setting cpu mode again

ekohl:
The value for `cpu.dig(:model, :name)` is always `nil` since that's the condition in the `if` statement. It was intended to still respect the cpu mode option passed in.

Fixes: 7a3a46b48fe7 ("Prefer CPU host pass through over explicit models")


evgeni (approved):

ekohl:
https://github.com/fog/fog-libvirt/pull/157 aims to fix RuboCop.

Issue #162, opened 2024-11-18
Ceph - Libvirt handle multiple pool


GaetanF:
Hello,

After checking the ceph block on to_xml function of server.rb file. It's not possible to handle multiple Ceph pool on the `/etc/foreman/ceph.conf` file.

https://github.com/fog/fog-libvirt/blob/238a8b857593922354b432fda7d235534e352098/lib/fog/libvirt/models/compute/server.rb#L349

It will possible to handle them by splitting using comma the configuration string entry and check with them like this : 
`if ceph_args && ceph_args["libvirt_ceph_pool"]&.split(",")&.include?(volume.pool_name)`

Regards

ekohl:
Since you already suggested the solution, would you be willing to submit a PR for this?

PR #167, opened 2024-12-23
Implement Ceph multiple pools and tests

GaetanF:
Hello,

Following the issue #162, I have implemented the capability to handle multiple Ceph RBD Pool by using a comma-separated list on the ceph.conf file.

Regarding the Libvirt documentation, the host tag must be in the source tag and not after. I have fix them also.

I have also implemented the tests to handle the Ceph part by using a mock of the File object.

Feel free if you have questions


lib/fog/libvirt/models/compute/server.rb
```diff
@@ -346,13 +346,13 @@ def to_xml
 
                 volumes.each_with_index do |volume, index|
                   target_device = "vd#{('a'..'z').to_a[index]}"
-                  if ceph_args && volume.pool_name.include?(ceph_args["libvirt_ceph_pool"])
+                  if ceph_args && ceph_args["libvirt_ceph_pool"]&.split(",")&.include?(volume.pool_name)
```
ekohl, line 349: I'd prefer to fix this up in `read_ceph_args` so you don't do this in a loop. Same below for the monitor. Perhaps also add in ` .map(&:strip)` so you can also use spaces between the values while you're at it.

Another thing is that `libvirt_ceph_pools` suggests it's singular while so perhaps store the result of that as `libvirt_ceph_pools` to indicate it's an array?

lib/fog/libvirt/models/compute/server.rb
```diff
@@ -363,7 +363,7 @@ def to_xml
                         end
                       end
 
-                      xml.target(:dev => target_device, :bus => args["bus_type"] == "virtio" ? "virtio" : "scsi")
+                      xml.target(:dev => target_device, :bus => ceph_args["bus_type"] == "virtio" ? "virtio" : "scsi")
```
ekohl, line 366: I merged this separate so it properly closed the issue. Doing 2 different things (fixing a bug and adding a new feature) in the same commit is something I prefer to avoid.

GaetanF: Ok, I haven't seen the fix when I have started to work on it. I will rebase the merge on it.

lib/fog/libvirt/models/compute/server.rb
GaetanF: Ok ! I have updated the branch to move the split and verify keys in the read_ceph_args.

tests/libvirt/models/compute/server_tests.rb
```diff
@@ -1,10 +1,50 @@
+RealFile = File
+class FakeFile < RealFile
+  def self.file?(path)
+    if path == '/etc/foreman/ceph.conf'
+      return true
+    else
+      return RealFile.file(path)
+    end
```
ekohl, line 8: To remove some RuboCop concerns, perhaps reduce it to
```suggestion
    path == '/etc/foreman/ceph.conf' || RealFile.file(path)
```

ekohl (commented):
I think it would be good to have some fallback code for compatibility. If `libvirt_ceph_pools` isn't present, check for `libvirt_ceph_pool` and present it as `libvirt_ceph_pools` in `ceph_args`. I'm not sure if there's a deprecation mechanism, but it would be good to consider that too.

lib/fog/libvirt/models/compute/server.rb
```diff
@@ -458,13 +458,20 @@ def read_ceph_args(path = "/etc/foreman/ceph.conf")
...
+          valid_keys = ["monitor", "port", "libvirt_ceph_pools", "auth_username", "auth_uuid", "bus_type"]
+          array_values = ["monitor", "libvirt_ceph_pools"]
+
           File.readlines(path).each do |line|
             pair = line.strip.split("=")
-            key = pair[0]
-            value = pair[1]
-            args[key] = value
+            key = pair[0].strip
+            if valid_keys.include?(key)
```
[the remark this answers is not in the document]
GaetanF: The configuration file can contains some comments or extra lines. So I think it's better to only import configuration line respecting a minimal schema.

lib/fog/libvirt/models/compute/server.rb
```diff
@@ -458,13 +458,20 @@ def read_ceph_args(path = "/etc/foreman/ceph.conf")
...
+              value = array_values.include?(key) ? pair[1].split(',').map(&:strip) : pair[1].strip
+              args[key] = value
+            end
           end
 
+          puts args
```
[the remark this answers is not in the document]
GaetanF: Oups. I will remove that :)

tests/libvirt/models/compute/server_tests.rb
GaetanF: Nice ! I will change that. I'm more used to developing in Python than in Ruby.

lib/fog/libvirt/models/compute/server.rb
```diff
@@ -470,7 +470,10 @@ def read_ceph_args(path = "/etc/foreman/ceph.conf")
             end
           end
 
-          puts args
+          if args.has_key?("libvirt_ceph_pool")
```
ekohl, line 473: I'd be tempted to only read `libvirt_ceph_pool` if `libvirt_ceph_pools` isn't set. The presence of `libvirt_ceph_pools` IMHO indicates a user has migrated to the newer format. Then only read `libvirt_ceph_pool` as a fallback. Would you agree?

GaetanF: Okay. I think it's more of a transitive way of managing both at the same time. But in the end, you're right.

PR #170, opened 2025-02-25
Update default video device to vga

ekohl:
For over a decade the cirrus driver has been considered outdated and the vga driver is preferred. A common alternative is qxl, but that implies the use of spice which isn't shipped in at least RHEL.

It leaves out the vram and relies on libvirt to set a sane default.

At this moment I haven't tested it, but it came out of https://github.com/fog/fog-libvirt/pull/71.

Link: https://www.kraxel.org/blog/2014/10/qemu-using-cirrus-considered-harmful/

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -558,7 +558,7 @@ module Fog
             :cpu                    => {},
             :hugepages              => false,
             :guest_agent            => true,
-            :video                  => {:type => "cirrus", :vram => 9216, :heads => 1},
+            :video                  => {:type => "vga", :heads => 1},
             :virtio_rng             => {},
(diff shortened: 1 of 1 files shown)


geemus (approved):
Thanks for tracking down the specifics, lgtm.

geemus:
Do we need to release this?

ekohl:
I didn't have a chance to test this yet so I'd appreciate doing that before we release

geemus:
@ekohl No rush from me, I just didn't want to drop the ball. I think you can release also, so feel free to do so whenever you are ready (or ping me if I need to do it).

nofaralfasi:
I checked, and everything works fine. I was able to create a new Libvirt host with different types of video, and everything worked as expected.

PR #172, opened 2025-05-11
Deal with missing volume_type

### LINKED ISSUE #169: Volume loses its format_type when saved
When creating a volume with fog-libvirt, the format_type is lost:

```
irb(main):004:0> compute = Fog::Compute.new(provider: :libvirt, libvirt_uri: "qemu+ssh://s098209.rz.babiel.com/system")
=> #<Fog::Libvirt::Compute::Real:5400 @uri=#<Fog::Libvirt::Util::URI:0x0000000109fd8068 @parsed_uri=#<URI::Generic qemu+ssh://s098209.rz....
irb(main):005:0> vol = compute.volumes.new
=> 
  <Fog::Libvirt::Compute::Volume
...

irb(main):006:0> vol.format_type
=> "raw"


irb(main):007:0> vol.save
=> 
  <Fog::Libvirt::Compute::Volume
    id="tqVRqA-UU7o-qOms-HaHd-e5hr-a4qi-Fz4hU7",
    pool_name="vg1",
    key="tqVRqA-UU7o-qOms-HaHd-e5hr-a4qi-Fz4hU7",
    name="fog-817509623589251",
    path="/dev/vg1/fog-817509623589251",
    capacity=10,
    allocation=1,
    owner=nil,
    group=nil,
    format_type=nil,
    backing_volume=nil
  >

irb(main):008:0> vol.format_type
=> nil
```

If this created volume is then used for a VM, the generated XML is invalid: 

```
<driver name="qemu" type=""/>
```

which leads to the error `Call to virDomainDefineXML failed: unsupported configuration: unknown driver format value ''` 

This causes https://projects.theforeman.org/issues/38081

ekohl:
LVM volumes do not have a volume type and after a reload the value is set to nil. 0277e9b42589836134763dd182fe6025bf8349df introduced that reload after saving. The mock driver doesn't properly show this behavior, which is why it was missed. A test is added to simulate the real world behavior.

Fixes: 0277e9b42589836134763dd182fe6025bf8349df ("Correctly load volumes after creation and cloning")
Link: https://libvirt.org/formatstorage.html#storage-volume-target-elements
Link: https://libvirt.org/storage.html#valid-logical-volume-format-types

Fixes #169

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -368,7 +368,9 @@ module Fog
                   else
                     is_block = volume.path.start_with?("/dev/")
                     xml.disk(:type => is_block ? "block" : "file", :device => "disk") do
-                      xml.driver(:name => "qemu", :type => volume.format_type)
+                      driver = xml.driver(:name => "qemu")
+                      driver[:type] = volume.format_type if volume.format_type
+
                       if is_block
                         xml.source(:dev => volume.path)
                       else
--- a/lib/fog/libvirt/models/compute/volume.rb
+++ b/lib/fog/libvirt/models/compute/volume.rb
@@ -91,7 +91,7 @@ module Fog
               xml.capacity(capacity_size, :unit => capacity_unit)
 
               xml.target do
-                xml.format(:type => format_type)
+                xml.format(:type => format_type) if format_type
                 xml_permissions(xml)
               end
 
--- a/tests/libvirt/models/compute/server_tests.rb
+++ b/tests/libvirt/models/compute/server_tests.rb
@@ -129,5 +129,39 @@ Shindo.tests('Fog::Compute[:libvirt] | server model', ['libvirt']) do
 
       os_firmware && secure_boot && enrolled_keys && loader_attributes
     end
+
+    test("with volumes") do
+      pool = Fog::Compute[:libvirt].pools.create(
+        :persistent => true,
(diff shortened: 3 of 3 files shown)


ekohl:
@nofaralfasi mind taking a look? I take https://community.theforeman.org/t/invalid-libvirt-xml-unknown-driver-format-value/43081/10 as a confirmation that it works.

mxey:
Handling a nil volume_type when creating the server might fix the issue with Foreman though, but it's IMO more of a workaround.

ekohl:
@mxey as indicated in the linked resources, this is mirroring what libvirt internally does. LVM doesn't have a format type so pretending it does doesn't make sense.

mxey:
> @mxey as indicated in the linked resources, this is mirroring what libvirt internally does. LVM doesn't have a format type so pretending it does doesn't make sense.

OK, then the default `format_type` set to `raw` also does not make sense. This PR fixes the problem if used through Foreman, but it does not solve the bug as I filed it, which affects more API. I personally do not care about the other API, but maybe other users of the library do?

ekohl:
I debated dropping the default of raw. Might still revisit that

lestercheung:
For the record we have deployed this fix since 2025 Q3 which works for us (Foreman v3.17) and I agree with @ekohl's interpretation that LVM volume doesn't have a format from libvirt's point of view and thus shouldn't have a format type set.

PR #185, opened 2026-01-08
Add Trusted Platform Module (TPM) support

rhjyoung:
Add Trusted Platform Module (TPM) support

Signed-off-by: Jerone Young jyoung@redhat.com

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -586,7 +586,7 @@ module Fog
             :video                  => {:type => "virtio", :heads => 1},
             :virtio_rng             => {},
             :firmware_features      => { "secure-boot" => "no" },
-            :tpm                    => {:enable => false, :id => "tpm0"}
+            :tpm                    => {:enable => false}
           }
         end
 
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -416,7 +416,6 @@ module Fog
 
                 if tpm[:enable]
                   tpm_model_type = (tpm_device.model == "spapr-tpm-proxy") ? "spapr-tpm-proxy" : "tpm-#{tpm_device.model}"
-                  
                   xml.tpm(:model => tpm_model_type) do
                     if tpm_device.type == "passthrough"
                       xml.backend(:type => tpm_device.type) do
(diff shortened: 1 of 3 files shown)


lib/fog/libvirt/models/compute/server.rb
```diff
@@ -561,6 +591,7 @@ def defaults
             :video                  => {:type => "virtio", :heads => 1},
             :virtio_rng             => {},
             :firmware_features      => { "secure-boot" => "no" },
+            :tpm                    => {:enable => false, :id => "tpm0"}
```
ekohl: I think the ID isn't that useful here because any user who overrides will need to specify it. Perhaps use:
```suggestion
            :tpm                    => false,
```

Then in `initialize_tpm` you can rewrite it and implement the default if it's set to `true`?

ekohl (commented):
These are just my first impressions. Haven't done actual testing on it.

rhjyoung:
Created code for Foreman:
        https://github.com/theforeman/foreman/pull/10824

ekohl (commented):
Code looks good to me. A quick test in `tests/libvirt/models/compute/server_tests.rb` tells me it works. I used this test:
```ruby
test("with tpm set to true") do
  server = Fog::Libvirt::Compute::Server.new(
    {
      :tpm => true,
      :nics => [],
      :volumes => []
    }
  )
  xml = server.to_xml

  xml.include?('<tpm model="tpm-crb">') && xml.include?('<backend type="emulator" version="2.0"/>')
end
```

Then get the following XML:
```xml
<tpm model="tpm-crb">
  <backend type="emulator" version="2.0"/>
</tpm>
```

Is this what you'd expect?

rhjyoung:
Yeap for x86_64 that is the expected result.

rhjyoung:
Anything else required for this to be merged ?

lib/fog/libvirt/models/compute/tpm.rb
```diff
@@ -0,0 +1,97 @@
...
+          @id = "tpm0"
+          @arch = arch
+          super defaults.merge(attributes)
+          raise Fog::Errors::Error, "#{type} is not a supported TPM type" if new? && !TYPES.include?(type)
+          raise Fog::Errors::Error, "#{model} is not a supported TPM model" if new? && !supported_models.include?(model)
+          raise Fog::Errors::Error, "TPM model type crb does not a supported TPM version 1.2" if model == "crb" && version == "1.2"
+        end
+
+        def new?
+          id.nil?
```
ekohl: If you assign an id in `initialize`, is this ever true?

PR #189, opened 2026-03-31
fix: handle nil network name for bridge/direct interfaces

ajmeese7:
## Problem

When a VM has a bridge or macvtap interface (e.g. from a Vagrant `public_network` with `:dev`), libvirt represents it without a source network attribute, so the nic's network name is `nil`. `addresses()` took `nics.first` unconditionally and passed that `nil` name down to `lookup_network_by_name(nil)`, raising:

```
TypeError: no implicit conversion of nil into String
```

This crashed `vagrant-libvirt` state checks on any VM with a bridge NIC.

## Fix

Primary fix is in `addresses()`: select the first nic that actually has a network instead of blindly taking the first nic. Bridge/direct interfaces have no network name and never carry a libvirt-managed lease, so skipping them both avoids the nil lookup and gives multi-nic VMs a better chance of resolving a real address.

As defense in depth, `get_network_by_filter` still returns `nil` early when the name filter is `nil`, and `list_networks` compacts the result so `Fog::Collection#load` never receives a nil entry. This keeps `lookup_network_by_name(nil)` from raising for any other caller. Happy to drop this layer if you'd rather keep the surface minimal.

## Tests

Added minitest coverage for `addresses()`:

- A bridge-only VM (single nic, nil network) returns no address and never performs a network lookup.
- A mixed VM (bridge nic first, NAT nic second) resolves the address from the networked nic.

Fixes `vagrant-libvirt` state checks crashing on VMs with bridge NICs.

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -478,7 +478,11 @@ module Fog
         # if the server has multiple network interface
         def addresses(service_arg=service, options={})
           ip_address = nil
-          if (nic = self.nics&.first)
+          # Pick the first nic that actually has a network. Bridge/direct
+          # interfaces carry a nil network name, which can't be looked up and
+          # never has a libvirt-managed lease, so skip them in favour of a nic
+          # that can resolve to an address.
+          if (nic = self.nics&.find { |n| n.network })
             net = service.networks.all(:name => nic.network).first
             # Assume the lease expiring last is the current IP address
             ip_address = net&.dhcp_leases(nic.mac)&.max_by { |lse| lse["expirytime"] }&.dig("ipaddr")
--- a/lib/fog/libvirt/requests/compute/list_networks.rb
+++ b/lib/fog/libvirt/requests/compute/list_networks.rb
@@ -9,7 +9,7 @@ module Fog
               data << network_to_attributes(client.lookup_network_by_name(network_name))
             end
           else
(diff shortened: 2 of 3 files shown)


lib/fog/libvirt/version.rb
```diff
@@ -1,5 +1,5 @@
 module Fog
   module Libvirt
-    VERSION = '0.15.0'
+    VERSION = '0.15.1'
```
ekohl: Please don't bump the version in this commit.

ekohl (changes requested):
Please address the RuboCop issue too

ekohl (commented):
> `fog-libvirt`'s `addresses()` passes `nic.network` (nil) to `list_networks`, which calls `lookup_network_by_name(nil)`, raising:

I'm rereading this now and this is:
https://github.com/fog/fog-libvirt/blob/eb3ca300667b5303e1f1f4ec645b058ff07963b5/lib/fog/libvirt/models/compute/server.rb#L475-L489
In particular:
https://github.com/fog/fog-libvirt/blob/eb3ca300667b5303e1f1f4ec645b058ff07963b5/lib/fog/libvirt/models/compute/server.rb#L482

Isn't the better fix to try and find the first nic with a network? So:
```ruby
if (nic = self.nics&.filter { |nic| nic.network }&.first)
```

That gives a better chance of finding a network that actually has an ip address.

ekohl (approved):

ekohl (approved):
Thanks!

PR #190, opened 2026-04-01
fix: fall back to lease file when DHCPLeases API is empty

ajmeese7:
## Summary

- When a libvirt network uses an external dnsmasq (not managed by libvirt), the `DHCPLeases` API returns empty and the VM's IP address can't be discovered. This adds a fallback that reads the dnsmasq lease file directly from `/var/lib/libvirt/dnsmasq/<network>.leases`.
- Common scenario: dnsmasq running inside a network namespace on WSL2 to work around port 67 conflicts with Windows ICS.
- Includes hardening (error handling for file permission/race conditions, warning log on fallback) and a minitest suite for the new method.

## Changes

- **`lib/fog/libvirt/models/compute/server.rb`**
  - New `ip_address_from_leasefile(net, mac)` private method parses dnsmasq lease files, matches by MAC, returns IP with latest expiry
  - `addresses()` calls the fallback when `dhcp_leases` returns nil and a network exists
  - `DNSMASQ_LEASE_DIR` constant for the lease file directory
  - `Fog::Logger.warning` when the fallback path is entered (emitted only once per MAC)
  - `Errno::EACCES`/`Errno::ENOENT` rescue around file reads
  - Removed unused `DOMAIN_CLEANUP_REGEXP` constant
  - Retain `Tempfile` reference in `generate_config_iso_in_dir` to prevent GC from deleting the ISO file before `File.size` reads it (fixes flaky CI `ENOENT` failure)
- **`minitests/server/leasefile_fallback_test.rb`** — 8 test cases covering happy path, expiry selection, case-insensitive MAC, missing file, malformed lines, no match, permission error, and nil network name

## Test plan

- [x] `bundle exec rake minitest` passes
- [x] Manual verification on a WSL2 host with external dnsmasq in a network namespace
- [x] Verify existing `public_ip_address` behavior is unchanged when libvirt DHCP is active (`bundle exec rake test`)

### DIFF
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -523,10 +523,10 @@ module Fog
             # external dnsmasq running in a network namespace on WSL2), the
             # DHCPLeases API returns empty.  Read the dnsmasq lease file directly.
             if ip_address.nil? && net
-              @leasefile_fallback_warned ||= {}
-              unless @leasefile_fallback_warned[nic.mac]
+              @@leasefile_fallback_warned ||= {}
+              unless @@leasefile_fallback_warned[nic.mac]
                 Fog::Logger.warning("DHCPLeases API returned no address for #{nic.mac}; falling back to dnsmasq lease file.")
-                @leasefile_fallback_warned[nic.mac] = true
+                @@leasefile_fallback_warned[nic.mac] = true
               end
               ip_address = ip_address_from_leasefile(net, nic.mac)
             end
--- a/lib/fog/libvirt/models/compute/server.rb
+++ b/lib/fog/libvirt/models/compute/server.rb
@@ -259,11 +259,14 @@ module Fog
           user_data_path = File.join(dir_path, 'user-data')
           File.write(user_data_path, user_data)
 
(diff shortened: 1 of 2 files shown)


lib/fog/libvirt/models/compute/server.rb
```diff
@@ -259,11 +259,14 @@ def generate_config_iso_in_dir(dir_path, user_data, &blk)
           user_data_path = File.join(dir_path, 'user-data')
           File.write(user_data_path, user_data)
 
-          isofile = Tempfile.new(['init', '.iso']).path
+          iso_tempfile = Tempfile.new(['init', '.iso'])
```
ekohl, line 262: These changes look completely unrelated, but still look good. Could you please split those into its own PR?

ekohl (commented):
In a networked setup this won't work and I'm not sure this is something I'd want to support. This gem is currently struggling for maintainers and all added code needs to be maintained. I'm inclined to reject this PR.

lib/fog/libvirt/models/compute/server.rb
ajmeese7: Handling this in #191

PR #196, opened 2026-07-27
Implement full support for network configuration

davispuh:
This PR implements full support for network configuration.
I basically implemented pretty much all of libvirt [Network](https://libvirt.org/formatnetwork.html) functionality.

Example usage:
```ruby
require "fog/libvirt"

compute = Fog::Compute.new(provider: "Libvirt", libvirt_uri: "qemu:///system")

# Create new network
network = compute.networks.new(name: "example",
                               forward: { mode: :nat },
                               dns: {
                                 hosts: [
                                   { ip: "192.168.25.25", hostnames: ["example.org"] },
                                 ],
                                 txts: [
                                   { name: "example", value: "some text" },
                                 ],
                               },
                               ips: [
                                 { address: "192.168.40.1", netmask: "255.255.255.0",
                                   dhcp: { ranges: [{ start: "192.168.40.100", end: "192.168.40.200" }] } },
                                 { address: "2001:db8::1", prefix: 64, family: :ipv6,
                                   dhcp: { ranges: [{ start: "2001:db8::100", end: "2001:db8::200" }] } },
                               ])
network.save

network.start
network.enable_autostart

# Modify existing
existing = compute.networks.all(:name => "example").first
existing.ips[0].dhcp.ranges << ("192.168.40.210".."192.168.40.230")
existing.dns.txts[0].value = "new val"
existing.dns.hosts = []

# these changes will be applied without destroy/recreate because libvirt supports that in some cases
# in other cases this will delete network and recreate with new changes (will have same name and uuid)
existing.save

existing.stop

# convert persistent network to transient
existing.persistent = false
existing.save

puts existing.to_xml

existing.destroy
```

I also added comprehensive tests that exercise all functionality. By specifying `LIBVIRT_URI=qemu:///session` env var you can run them against actual libvirt instance (`qemu:///system` also works)

### DIFF
--- a/lib/fog/libvirt/models/compute/network/nat.rb
+++ b/lib/fog/libvirt/models/compute/network/nat.rb
@@ -1,0 +1,63 @@
+require "fog/core/model"
+require_relative "../attribute_model"
+require_relative "../util/util"
+
+module Fog
+  module Libvirt
+    class Compute
+      class Network < Fog::Model
+        class Nat < Fog::Libvirt::Compute::AttributeModel
+          include Fog::Libvirt::Util
+
+          attribute :ipv6
+
+          attribute :address
+          attribute :port
+
+          remove_method :address=
+
+          def address=(address)
+            attributes[:address] = if address.is_a?(Range)
+                                     normalize_ip_range(address)
+                                   else
+                                     address
+                                   end
+          end
+
+          remove_method :port=
+
+          def port=(port)
+            if port.is_a?(Range)
+              attributes[:port] = normalize_number_range(port)
+            else
+              attributes[:port] = port
+              attributes[:port][:start] = attributes[:port][:start].to_i unless attributes[:port].to_h[:start].to_s.empty?
+              attributes[:port][:end] = attributes[:port][:end].to_i unless attributes[:port].to_h[:end].to_s.empty?
+            end
+          end
+
+          def self.parse_xml(node)
+            return nil unless node
+
+            attrs = xml_attrs(node)
+
+            attrs[:address] = xml_attrs(node.at_xpath("address"))
+            attrs.delete(:address) if attrs[:address].empty?
+
+            attrs[:port] = xml_attrs(node.at_xpath("port"))
+            attrs.delete(:port) if attrs[:port].empty?
+
+            attrs
+          end
+
+          def build_xml(xml)
+            xml.nat(attrs_xml(hash_except(attributes, :address, :port)).compact) do
+              xml.address(address.compact) if address
+              xml.port(port.compact) if port
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/fog/libvirt/models/compute/network/route.rb
+++ b/lib/fog/libvirt/models/compute/network/route.rb
@@ -1,0 +1,49 @@
+require "fog/core/model"
+require_relative "../util/util"
+
+module Fog
+  module Libvirt
+    class Compute
+      class Network < Fog::Model
+        class Route < Fog::Model
+          include Fog::Libvirt::Util
+
+          identity :gateway
+          attribute :address
+          attribute :netmask
+          attribute :prefix
+          attribute :family
+          attribute :metric
+
+          def initialize(attributes = {})
+            super(normalize_attrs(attributes))
+          end
+
+          def self.parse_xml(node)
+            return nil unless node
+
+            attrs = xml_attrs(node)
+            attrs[:prefix] = attrs[:prefix].to_i if attrs.key?(:prefix)
+            attrs[:metric] = attrs[:metric].to_i if attrs.key?(:metric)
+            attrs
+          end
+
+          def build_xml(xml)
+            xml.route(attributes.compact.transform_values(&:to_s))
+          end
+
+          private
+
+          def normalize_attrs(attrs)
+            attrs = { :gateway => attrs } if attrs.is_a?(String)
+            attrs[:prefix] = attrs[:prefix].to_i unless attrs[:prefix].to_s.empty?
+            attrs[:prefix] = attrs.delete("prefix").to_i unless attrs["prefix"].to_s.empty?
+            attrs[:metric] = attrs[:metric].to_i unless attrs[:metric].to_s.empty?
+            attrs[:metric] = attrs.delete("metric").to_i unless attrs["metric"].to_s.empty?
+            attrs
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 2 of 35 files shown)


lib/fog/libvirt/models/compute/network/nat.rb
```diff
@@ -0,0 +1,62 @@
...
+
+          def port=(port)
+            if port.is_a?(Range)
+              attributes[:port] = normalize_number_range(port)
+            else
+              attributes[:port] = port
+              attributes[:port][:start] = attributes[:port][:start].to_i unless attributes[:port].to_h[:start].to_s.empty?
+              attributes[:port][:end] = attributes[:port][:end].to_i unless attributes[:port].to_h[:end].to_s.empty?
+            end
+          end
```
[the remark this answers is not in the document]
davispuh: Copilot is wrong here. When setting `nat.port = nil` then `attributes[:port].to_h[:start].to_s.empty?` will evaluate to true and `attributes[:port][:start]` won't be touched.

lib/fog/libvirt/models/compute/network/route.rb
```diff
@@ -0,0 +1,48 @@
...
+
+          private
+
+          def normalize_attrs(attrs)
+            attrs = { gateway: attrs } if attrs.is_a?(String)
+            attrs[:prefix] = attrs[:prefix].to_i unless attrs[:prefix].to_s.empty?
+            attrs[:prefix] = attrs.delete("prefix").to_i unless attrs["prefix"].to_s.empty?
+            attrs[:metric] = attrs[:metric].to_i unless attrs[:metric].to_s.empty?
+            attrs[:metric] = attrs.delete("metric").to_i unless attrs["metric"].to_s.empty?
+            attrs
```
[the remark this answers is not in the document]
davispuh: Copilot is wrong here. `Route` inherits from `Fog::Model` so all attributes that will be passed to it will be set correctly. The reason why we need to check here for both is because we do casting to int before that so at this point we do need to convert both.

lib/fog/libvirt/models/compute/util/util.rb
```diff
@@ -17,6 +17,144 @@ def xml_elements(xml, path, attribute=nil)
...
+        attrs = attrs.to_h.transform_keys { |name| attr_camelcase(name).to_sym }
+        attrs.transform_values! { |value| value_xml(value) }
+        attrs
+      end
+
+      def normalize_ip_range(range)
+        end_val = range.end.to_s
+        if range.exclude_end?
+          last_ip = IPAddr.new(end_val)
+          end_val = IPAddr.new(last_ip.to_i - 1, last_ip.family).to_s
```
ekohl, line 141: Isn't this `last_ip - 1` really IPv4 logic and you can ignore that for IPv6?

minitests/network/network_test.rb
```diff
@@ -0,0 +1,685 @@
+require_relative "../test_helper"
+require "securerandom"
+require "nokogiri"
+require "json"
+require "ipaddr"
+
+class NetworkTest < Minitest::Test
+  LIBVIRT_URI = ENV["LIBVIRT_URI"] || "test:///default"
```
ekohl: IMHO this belongs in the helper that now now hardcodes the URI but I wouldn't mind if that changes to `ENV.fetch("LIBVIRT_URI", "test:///default")`:
https://github.com/fog/fog-libvirt/blob/ffe3e3b6cf3415998167721c1e0cd0883c50de46/minitests/test_helper.rb#L21-L23

minitests/network/network_test.rb
```diff
@@ -0,0 +1,685 @@
...
+  REAL_LIBVIRT = LIBVIRT_URI.start_with?("qemu")
+
+  def setup
+    @created_networks = []
+    if REAL_LIBVIRT
+      was_mocked = Fog.mock?
```
ekohl: Isn't this always true since the helper hardcodes mocking?
https://github.com/fog/fog-libvirt/blob/ffe3e3b6cf3415998167721c1e0cd0883c50de46/minitests/test_helper.rb#L20
That should respect the `FOG_MOCK` env var.

ekohl (changes requested):
This is a very large PR. I'd appreciate splitting it in 2 parts: first rewrite the existing network tests from shindo to minitest. Then the big addition of networking. That way we can first work out the test setup part and it's easier to verify the existing tests are rewritten.

lib/fog/libvirt/models/compute/util/util.rb
davispuh: No, this works correctly for IPv6. Purpose of this is to support excluding ranges, like consider:
```ruby
dhcp.ranges << ("2001:db8::1000"..."2001:db8::a000") # note 3 dots ...
```
This means wanted range is to `2001:db8::9fff` and this implements that correctly.

minitests/network/network_test.rb
davispuh: That would be better indeed. Will submit separate PR for that.

minitests/network/network_test.rb
davispuh: Yeah currently it's always true, but my thoughts are that when `LIBVIRT_URI` is set to `qemu` then intention is to run it against actual libvirt and not use mocking (thus ignoring FOG_MOCK). In my opinion having to set both `FOG_MOCK` and `LIBVIRT_URI` doesn't seem that good.
So I think we can implement it like this:
```ruby
mock = ENV.fetch("FOG_MOCK", "true")
mock = "false" if ENV['FOG_MOCK'].to_s.empty? && LIBVIRT_URI.start_with?("qemu")
Fog.mock! unless mock == "false"
```
And then don't specify ENV['FOG_MOCK'] anywhere in test files.
This will allow user to choose:
1. Not specifying anything => mocking with `test:///default`
2. Setting only LIBVIRT_URI => no mocks if qemu
3. Use FOG_MOCK to override

Also btw I noticed that currently FOG_MOCK=false is broken anyway and doesn't work.

ekohl: Your suggestion make sense. Minor bikeshedding: is `LIBVIRT_URI` as an env var is perhaps too generic and could leak in from other projects? Perhaps `FOG_LIBVIRT_URI` is better.

Then reading your code I think you can simplify your code using the block mode of `ENV.fetch`:
```ruby
mock = ENV.fetch("FOG_MOCK") { !LIBVIRT_URI.start_with?("qemu").to_s }
```

But your code also handles empty strings which may be useful because they're they easiest way to unset env vars.

ekohl:
Not a real review, but 1 thing that stood out to me:

> 4\. Fixed issues with Ruby 2.7 in CI (maybe drop it?). Had to add `hash_except` in Util because I can't live without `#except()` which is Ruby 3.0+

I think we could move to Ruby 3.0+ as well.

PR #197, opened 2026-07-28
Implement support for tests using real libvirt with FOG_LIBVIRT_URI

davispuh:
Currently `rake live` fails with:
```
Fog::Compute[:libvirt] (libvirt)     
    fog-libvirt/tests/libvirt/compute_tests.rb
      Fog::Compute[:libvirt] (libvirt)
    Missing required arguments: libvirt_uri (ArgumentError)
```
First commit from this PR fixes that so `rake live` works.
2nd commit implements support for running tests against real libvirt by using `FOG_LIBVIRT_URI` env var.
All these are fully supported and work correctly:
```
$ rake
$ FOG_MOCK=true rake

$ FOG_MOCK=true FOG_LIBVIRT_URI=qemu:///session rake
$ FOG_MOCK=true FOG_LIBVIRT_URI=qemu:///system rake

$ FOG_MOCK=false rake
$ rake live

$ rake live_session
$ FOG_LIBVIRT_URI=qemu:///session rake
$ FOG_MOCK=false FOG_LIBVIRT_URI=qemu:///session rake
$ FOG_LIBVIRT_URI=qemu:///system rake
$ FOG_MOCK=false FOG_LIBVIRT_URI=qemu:///system rake
```
(something like `FOG_LIBVIRT_URI=qemu+ssh://host/system rake` should also work but I haven't tested)

I also fixed tests so they cleanup afterwards so there won't be test leftovers when running against real libvirt.
But note that with this feature tests shouldn't do something like `servers.all.each(&:destroy)` to not cause someone a bad day when run with real libvirt. If needed then such should only be run when using libvirt test driver only.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -47,8 +47,7 @@ end
 
 desc 'Run tests'
 task :test do
-  mock = ENV['FOG_MOCK'] || 'true'
-  sh("export FOG_MOCK=#{mock} && bundle exec shindont tests")
+  sh("bundle exec shindont tests")
 end
 
 desc 'Run mocked tests'
@@ -59,6 +58,11 @@ end
 desc 'Run live tests'
 task :live do
   sh("export FOG_MOCK=false && bundle exec shindont tests")
+end
+
+desc 'Run live libvirt qemu:///session tests'
+task :live_session do
+  sh("export FOG_LIBVIRT_URI=qemu:///session && bundle exec rake test minitest")
 end
 
 desc "Open an irb session preloaded with this library"
--- a/minitests/test_helper.rb
+++ b/minitests/test_helper.rb
@@ -17,7 +17,11 @@ FileUtils.rm_f Dir.glob 'test/tmp/*.tmp'
 
 require 'fog/libvirt'
 
-Fog.mock!
-Fog.credentials = {
-    :libvirt_uri => 'test:///default',
-}.merge(Fog.credentials)
+Fog.credentials[:libvirt_uri] = ENV.fetch("FOG_LIBVIRT_URI", "test:///default")
+
+def real_libvirt?
+  Fog.credentials[:libvirt_uri].start_with?("qemu")
+end
+
+enable_mocking = ENV.fetch("FOG_MOCK") { (!real_libvirt?).to_s }
+Fog.mock! unless ["false", "no", "0"].include?(enable_mocking)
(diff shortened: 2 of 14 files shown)


tests/libvirt/models/compute/server_tests.rb
```diff
@@ -66,7 +66,9 @@
     end
 
     test('can destroy') do
-      servers.create(:name => Fog::Mock.random_letters(8)).destroy
+      server_name = "fog-test-destroy-#{Fog::Mock.random_letters(8)}"
+      servers.create(:name => server_name).destroy(:destroy_volumes => true)
+      servers.all.find { |server| server.name == server_name }.nil?
```
ekohl: You can use [Array#none?](https://rubyapi.org/o/array#method-i-none-3F) here to be even more expressive, right?
```suggestion
      servers.all.none? { |server| server.name == server_name }
```

minitests/test_helper.rb
```diff
@@ -17,7 +17,11 @@
...
-Fog.credentials = {
-    :libvirt_uri => 'test:///default',
-}.merge(Fog.credentials)
+Fog.credentials[:libvirt_uri] = ENV.fetch("FOG_LIBVIRT_URI", "test:///default")
+
+def real_libvirt?
+  Fog.credentials[:libvirt_uri].start_with?("qemu")
+end
+
+mock = ENV.fetch("FOG_MOCK") { (!real_libvirt?).to_s }
```
ekohl: Won't this turn into a global variable this way? Not sure if that is a problem though.

Rakefile
```diff
@@ -61,6 +60,11 @@ task :live do
   sh("export FOG_MOCK=false && bundle exec shindont tests")
 end
 
+desc 'Run live libvirt qemu:///session tests'
+task :live_session do
+  sh("export FOG_LIBVIRT_URI=qemu:///session && bundle exec rake test && bundle exec rake minitest")
```
ekohl: Rake can execute multiple tasks in the same invocation. Is that a benefit here or did you choose to use 2 separate invocations for separation?

```suggestion
  sh("export FOG_LIBVIRT_URI=qemu:///session && bundle exec rake test minitest")
```

ekohl (commented):
Thanks, looks really good. Only some minor inline comments but I'd be happy to merge this as is already.

tests/libvirt/models/compute/server_tests.rb
davispuh: Yep, applied.

minitests/test_helper.rb
davispuh: Nope, tested that helper variables aren't accessible in tests. But I changed name anyway to `enable_mocking`

Rakefile
davispuh: No reason, fixed.

ekohl (approved):
Thanks!

PR #198, opened 2026-07-29
Convert network_tests.rb to use Minitest

davispuh:
This PR converts `network_tests.rb` from Shindo to Minitest


minitests/network/network_test.rb
```diff
@@ -0,0 +1,36 @@
...
+    assert_kind_of Array, @network.dhcp_leases("aa:bb:cc:dd:ee:ff", 0) if Fog.mock?
+
+    attributes = [:name, :uuid, :bridge_name]
+    attributes.each do |attribute|
+      assert @network.respond_to? attribute
+    end
+
+    @network.attributes.each do |attribute|
+      @network.attributes.key? attribute
+    end
```
ekohl: To match the previous behavior this needs to be:
```suggestion
    attributes.each do |attribute|
      assert @network.attributes.key? attribute
    end
```

ekohl (changes requested):

minitests/network/network_test.rb
davispuh: Indeed, just pushed fix.

ekohl (approved):

